Issue #20, opened 2019-05-24
Use JupyterHub for REST api authentication


jcrist:
Currently we support REST api authentication via:

- basic (mostly for testing)
- kerberos

We could also handle REST api authentication via validating a JupyterHub api token with JupyterHub. This would allow nice integration with JupyterHub, which provides an api token in the kernel environment.

jcrist:
These docs look useful: https://jupyterhub.readthedocs.io/en/stable/api/services.auth.html

jcrist:
Fixed by #36.

PR #36, opened 2019-06-04
Add JupyterHub integration

### LINKED ISSUE #23: Integrate as JupyterHub service
This is separate but related to #20. With JupyterHub for authentication we should be able to integrate nicely as a [JupyterHub service](https://jupyterhub.readthedocs.io/en/stable/getting-started/services-basics.html). It would be good to get this working, and document the deployment steps involved.

jcrist:
- Adds support for using JupyterHub for authentication
- Adds support for running behind JupyterHub's proxy as a managed
service.

Fixes #23, #20.

### DIFF
--- a/continuous_integration/before_install.sh
+++ b/continuous_integration/before_install.sh
@@ -1,3 +1,5 @@
 gimme 1.12 \
 && source ~/.gimme/envs/go1.12.env \
-&& go get github.com/stretchr/testify/assert
+&& go get github.com/stretchr/testify/assert \
+&& nvm install 6 \
+&& nvm use 6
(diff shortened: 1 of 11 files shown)


jcrist:
This allows users to login to JupyterHub, and from inside a notebook session connect to the `dask-gateway` transparently (the configuration values in the `Gateway(...)` call below can be set in the dask config, so `gateway = Gateway()` is possible).

Screen Shot 2019-06-04 at 12 26 11 PM

It also allows JupyterHub to deploy dask-gateway as a managed service, which could be nice for simple deployments.

jcrist:
cc @mrocklin, you may be interested in this.

mrocklin:
cc @jacobtomlinson @hhuuggoo @jhamman

PR #43, opened 2019-06-18
Add kubernetes cluster manager

### LINKED ISSUE #12: Add Kubernetes support
We currently support yarn and local processes, it would be good to also add support for Kubernetes. See [dask-kubernetes](https://github.com/dask/dask-kubernetes) for reference.

jcrist:
Adds initial kubernetes support as a backend.

Most of the work here was getting the test setup working (minikube is sufficiently different than running in a docker container that lots of extra work was needed). The actual kubernetes backend code is fairly short and should hopefully be clear.

Fixes #12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,18 @@ jobs:
           script:
             - docker exec master /working/continuous_integration/docker/hadoop/script.sh
 
+        - if: (commit_message !~ skip-tests) AND (commit_message ~= test-kube)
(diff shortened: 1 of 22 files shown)


mrocklin:
Pinging @jhamman, @quasiben, and @jacobtomlinson , just so that they see this when they resurface.

I wouldn't wait on them though, they're all away from work at the moment.

Issue #41, opened 2019-06-10
Refactor Cluster/Worker states to reduce polling frequency on startup


jcrist:
Currently clusters/workers go through the following states on startup:

- `STARTING`: Waiting on the `ClusterManager.start_*` method
- `STARTED`: `ClusterManager.start_*` has returned, waiting for the process to connect to the gateway
- `RUNNING`: Process has connected, and is deemed functional.

To quickly catch failures (rather than just timeout on process startup), cluster managers are responsible for tracking process state until it's deemed STARTED. For some cluster backends this can be expensive, and would be good to avoid. During perfect execution (no failures), tracking job status isn't even necessary, it's only really useful for getting quick feedback on failures.

As such, it would be good to split. I propose the following states:

- `SUBMITTING`: Waiting on the `ClusterManager.submit_cluster`/`ClusterManager.submit_worker` method. This can return as soon as some other system is responsible for managing the process (e.g. YARN/kerberos/job queues.
- `STARTING`: `ClusterManager.submit_*` has returned.
- `STARTED`: Process is deemed started (more on this later)
- `RUNNING`: Process has connected, and is deemed functional.

Once a process (cluster or worker) is in the `STARTING` state, the gateway starts a task to wait for the process to start (which could use polling or something more efficient, depending on the cluster backend). If polling, this period should be configurable, and should be longer than the current one (we want it small enough that an interactive user won't grow impatient on debugging failing jobs, but large enough that it won't take too much time. 15-30 seconds could be a good default). The method could be called `wait_for_cluster`/`wait_for_worker` or something like that.

If the waiting task returns true, the process will be moved to the `STARTED` state, and a timeout will be started for the process to connect. If the process connects before it's moved to the `STARTED` state (with a large polling interval this isn't unlikely), the waiting task can be cancelled, and the process can move directly to `RUNNING`. If the waiting task times out, the normal timeout logic is applied.

For cluster backends that have no "submit" concept (such as local processes), the `wait_for_*` methods would be no-ops.

---

The reason we can't decrease the polling frequency with the current design is that `start_*` tasks block until the process is started, and can't be cancelled for a successful startup (since they're responsible for returning any process state to be saved). This means that if a process connects before the `start_*` method returns we still have to wait for the `start_*` method to return before moving the process into the `RUNNING` state. By splitting the state into `SUBMITTING` and `STARTING`, we divide a task that may return state that needs saving (e.g. `job_id`), and a task that's just waiting on a job and may be cancelled.

PR #49, opened 2019-06-28
Add `worker_status` checks

jcrist:
Similar to `ClusterManager.cluster_status`, this adds
`ClusterManager.worker_status` for checking the status of a worker. This
allows detecting failed workers early (rather than waiting for
connection to timeout), while not consuming too many resources.

Second part of answering #41.


jcrist:
Hmmm, the kubernetes tests pass locally but not on travis. Debugging now.

PR #59, opened 2019-07-25
[WIP] Helm Chart

### LINKED ISSUE #44: Helm Chart for kubernetes deployment
As of #43 we support kubernetes as a backend. It would be good to develop a helm chart for deploying this on kubernetes clusters. I have no helm experience, but may eventually be able to figure this out - others will likely have an easier time at it. Here's a high level of how I see a kubernetes deployment looking:

- The gateway will have three components running and managed by a deployment:
  
  - The gateway process (`dask-gateway-server`)
  - The scheduler proxy (`dask-gateway-server scheduler-proxy`)
  - The web proxy (`dask-gateway-server web-proxy`)

  The resources for these are all part of the `dask-gateway-server` python package, and could all be run using the same image (but in different containers). I think we'd want different containers rather than running in the same container to let kubernetes handle restarts (the gateway should be robust to transient component restarts). The appropriate configuration would need to be set to allow for this (this exists, but is only documented at the traitlets config level).

- We'd want to expose configuration for common things via helm. We should be able to do this similar to the way jupyterhub does (https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/images/hub/jupyterhub_config.py). A few things people may be interested in:

  - All the configurable traitlets on `dask_gateway_server.managers.kubernetes.KubeClusterManager`
  - The authenticator used
  - The database url
  - Various timeouts

- I'm not sure how people would run this with a persistent database in kubernetes. A few ideas:

  - In memory database only, not robust to gateway restarts (this should maybe be the default)
  - Sqlite stored on a PVC
  - Some custom db url (the user would be responsible for setting up the db externally)

- We should provide default images to use. One already exists for running the scheduler/worker processes [here](https://github.com/jcrist/dask-gateway/tree/master/resources/docker/dask-gateway). All deployment resources should go somewhere under [this directory](https://github.com/jcrist/dask-gateway/tree/master/resources).

- It would be good to have a way to demo this locally on minikube. We have a demo for a hadoop cluster [here](https://github.com/jcrist/dask-gateway/tree/master/demo), it would be nice to replicate this for kubernetes.

jcrist:
This adds a helm chart for deploying on Kubernetes. The work here is inspired by that of [zero-to-jupyterhub-k8s](https://github.com/jupyterhub/zero-to-jupyterhub-k8s), with some changes and simplifications to suite our needs.

Fixes #44


jcrist:
This still needs some work, but I'm able to make requests from outside the cluster, and everything seems to be communicating fine. I'm very new at using helm, so there's likely some stuff I'm doing wrong. Hopefully will have time tomorrow to keep poking at things.

jcrist:
cc @quasiben, this may interest you.

mrocklin:
cc @jacobtomlinson

resources/helm/dask-gateway/templates/scheduler-proxy-service.yaml
```diff
@@ -0,0 +1,32 @@
...
+    app.kubernetes.io/component: scheduler-proxy
+spec:
+  selector:
+    {{- include "dask-gateway.labels" . | nindent 4 }}
+    app.kubernetes.io/component: scheduler-proxy
+  ports:
+    - protocol: TCP
+      port: 8786
+      targetPort: 8786
+  type: LoadBalancer
```
jacobtomlinson: I would make this configurable. Load balancers can be expensive and people often consolidate those costs using ingresses.

jcrist: This service uses raw `tcp` and not `http`, which as I understand doesn't work with kubernetes ingress? The other proxy definitely could use an ingress.

jacobtomlinson: You're right that reverse proxies are layer 7 and therefore only support http/https. 

However some ingress controllers like nginx [do support tcp forwarding](https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services/). They actually avoid nginx all together and just add a new port endpoint to the existing loadbalancer and forward the traffic straight into your service. It requires a little more setup on the user's part, they would need to create an additional resource as well as install the helm chart, but they would need to be able to set the type to `ClusterIP` in the chart.

jacobtomlinson: [Looks like trafik also supports TCP 🎉.](https://github.com/containous/traefik/pull/4587)

jcrist: I've made the type configurable, defaulting to `LoadBalancer` (mirroring that done in zero-to-jupyterhub-k8s).

jcrist:
Ok, I'm going to merge this as is. I've confirmed everything works fine in minikube, and think I've followed most of the best practices as defined in:

- https://github.com/helm/helm/blob/master/docs/chart_best_practices
- https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md

Further updates can be done later. I don't currently have a setup for testing on a cloud (I also don't have any experience with that, but will eventually figure it out).

mrocklin:
cc also @jhamman who is also doing helm work over at Pangeo, and so may be
in a good frame of mind to take a look at this right now

(one of the benefits of subscribing to too many issue trackers I guess)

PR #65, opened 2019-07-29
Better errors on import failure

### LINKED ISSUE #58: Better error messages for missing deps
When trying to use a cluster manager with uninstalled external dependencies the error message is a bit too general.  For example, when trying to use `KubeClusterManager` I get

> The 'cluster_manager_class' trait of <dask_gateway_server.app.DaskGateway object at 0x7ff2ba00f198> instance must be a type, but 'dask_gateway_server.managers.kubernetes.KubeClusterManager' could not be imported

Installing the kubernetes python library (pip install kubernetes) resolves the issue

jcrist:
We override the ``Type`` traitlet to provide a better error message on
import error, and raise nice error messages for cluster managers that
have extra dependencies.

Fixes #58.


quasiben:
👍

PR #66, opened 2019-07-30
Catch HTTPError from timeouts in client

jcrist:
With tornado 6, `HTTPError`s from timeouts are no longer returned
(and are instead raised) even with `raise_error=False`. We now catch
these and add a test for this.


jcrist:
cc @ericdill.

ericdill:
thx @jcrist -- testing this now

PR #67, opened 2019-07-31
Cleanup client-side exceptions

jcrist:
Switches to using specific error classes for client errors instead of
raw `Exception`.

Also fixes an issue where errors during cluster startup were being conflated
with `connect` long polling timeouts, resulting in poor exceptions
client-side. A test has been added for this code path.


jcrist:
cc @ericdill, this should fix the json decode exception you were seeing while blocking on `new_cluster`.

ericdill:
I can confirm that this fixes the stack trace I was getting. Clusters are successfully starting and I can run simple dask code against it
Jul 31-15 09 06

PR #70, opened 2019-07-31
Add _repr_html_ to cluster object

jcrist:
Adds support for standard HTML repr for the case where ipywidgets is not
installed.

If ipywidgets is installed, notebook reprs use the standard widgets. If not, notebook reprs look like the following:

- Dashboard running

Screen Shot 2019-07-31 at 6 14 57 PM

- No dashboard running

Screen Shot 2019-07-31 at 6 16 46 PM

This provides a nicer user experience for users without ipywidgets.


jcrist:
cc @ericdill, this should make use in AE5 nicer

PR #78, opened 2019-09-05
Change defaults for easier local experience

jcrist:
This changes the default configuration so that it "just works" when run
locally for demos/experimentation, without need for any additional
configuration.

- Default cluster manager is `UnsafeLocalClusterManager`
- Default authenticator is `DummyAuthenticator`

Note that both of these choices are bad/unsafe for real-world usage, but
any real world deployment will also need to configure other options
anyway.

Users can now try out dask gateway by doing:

```console
$ pip install dask-gateway dask-gateway-server
$ dask-gateway-server
```

And in another terminal:

```python
import dask_gateway
gateway = dask_gateway.Gateway(<address-logged-above>)
```

---

This also fixes a bug in our default configuration, where the hostname
wouldn't be resolved (since traitlets doesn't run validators on default
values).

### DIFF
--- a/dask-gateway-server/dask_gateway_server/app.py
+++ b/dask-gateway-server/dask_gateway_server/app.py
@@ -137,14 +137,14 @@ class DaskGateway(Application):
     )
 
     authenticator_class = Type(
-        "dask_gateway_server.auth.KerberosAuthenticator",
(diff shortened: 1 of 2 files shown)


mrocklin:
Woot.  I know what I'm doing on my next plane ride :)

PR #85, opened 2019-09-10
A few URL cleanups

jcrist:
This fixes a few things:
- When specified to listen on all interfaces, we previously accidentally resolved to a single ip before listening. We now pass through the original URL unchanged.
- We now offset the URLs that matter in the logs from those that are just logged for debugging. This should make it clearer for users what they need to actually connect.
- For addresses listening on all interfaces, we log both the hostname and localhost as options to connect.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/app.py
+++ b/dask-gateway-server/dask_gateway_server/app.py
@@ -7,7 +7,7 @@ import stat
 import tempfile
 import weakref
 from functools import partial
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urlparse
 
 from tornado import web
 from tornado.log import LogFormatter
@@ -31,7 +31,7 @@ from .objects import (
 )
 from .options import Options
 from .proxy import SchedulerProxy, WebProxy
-from .utils import cleanup_tmpdir, cancel_task, TaskPool, Type, get_ip
+from .utils import cleanup_tmpdir, cancel_task, TaskPool, Type, get_connect_urls
 
 
 # Override default values for logging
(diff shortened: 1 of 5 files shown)


jcrist:
Running with no configuration, the logs now look like:

```
$ dask-gateway-server
[I 2019-09-10 11:44:13.152 DaskGateway] Generating new cookie secret
[I 2019-09-10 11:44:13.154 DaskGateway] Generating new auth token for scheduler proxy
[I 2019-09-10 11:44:13.154 DaskGateway] Starting the Dask gateway scheduler proxy...
[I 2019-09-10 11:44:13.156 DaskGateway] Dask gateway scheduler proxy started at 'tls://:8786', api at 'http://127.0.0.1:55997'
[I 2019-09-10 11:44:13.163 DaskGateway] Generating new auth token for web proxy
[I 2019-09-10 11:44:13.164 DaskGateway] Starting the Dask gateway web proxy...
[I 2019-09-10 11:44:13.166 DaskGateway] Dask gateway web proxy started at 'http://:8000', api at 'http://127.0.0.1:56002'
[I 2019-09-10 11:44:13.276 DaskGateway] Gateway private API listening at http://127.0.0.1:8081
[I 2019-09-10 11:44:13.282 DaskGateway] Dask-Gateway started successfully!
[I 2019-09-10 11:44:13.282 DaskGateway] - Public address at http://127.0.0.1:8000 or http://0460-jcrist:8000
[I 2019-09-10 11:44:13.282 DaskGateway] - Proxy address at tls://127.0.0.1:8786 or tls://0460-jcrist:8786
```

jcrist:
cc @mrocklin ^^

mrocklin:
Cool.  That's super clear.  If we wanted to be more idiot-proof, we might highlight the address more with `----------------` lines or something.  We might also want the most used address to be last.  

However, this might not be that important.  Presumably most naive users aren't looking at these logs, so probably it's not worth a bunch of time to micro-optimize this.

jcrist:
Yeah, these will only be seen by administrators or people trying it out locally. We want it to be nice to use locally for the second group, but the first group will be more common in practice.

PR #93, opened 2019-09-16
[WIP] Add support for user limits

### LINKED ISSUE #25: User restrictions
From an admin perspective it may sometimes be useful to support restrictions on user resource usage. A few examples:

- Maximum number of workers per user
- Maximum resources per user
- Maximum number of clusters per user

We should probably provide a common metrics, but will likely also need a generic hook for operations an admin may want to restrict. This may look like:

```python
def can_start_cluster(user):
    """Check if a user can start a cluster.
    
    Returns
    -------
    allowed : bool
        True if the user can start a cluster, False otherwise
    msg : str or None
        A user-facing message explaining why the operation is currently forbidden.
    """
    pass

def can_start_worker(user, cluster):
    """Check if a user can start a worker.
    
    Returns
    -------
    allowed : bool
        True if the user can start a worker, False otherwise
    msg : str or None
        A user-facing message explaining why the operation is currently forbidden.
    """
    pass
```

jcrist:
Adds support for server-side limits of the following:
- Total cores per user
- Total memory per user
- Total clusters per user

From a user, hitting the limits results in:
- An informative error if creating a new cluster would exceed a limit
- A warning if scaling to ``n`` workers would exceed a limit

Fixes #25.

Still needs tests.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/app.py
+++ b/dask-gateway-server/dask_gateway_server/app.py
@@ -29,6 +29,7 @@ from .objects import (
     normalize_encrypt_key,
     is_in_memory_db,
 )
+from .limits import UserLimits
 from .options import Options
 from .proxy import SchedulerProxy, WebProxy
 from .utils import (
@@ -416,6 +417,7 @@ class DaskGateway(Application):
         self.init_scheduler_proxy()
         self.init_web_proxy()
         self.init_authenticator()
+        self.init_user_limits()
         self.init_database()
         self.init_tornado_application()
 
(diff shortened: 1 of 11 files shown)


jcrist:
From a admin perspective, adding limits looks like:

```python
# Each user is limited to no more than 50 cores
c.UserLimits.max_cores = 50

# Each user is limited to no more than 100 GiB
c.UserLimits.max_memory = "100 GiB"

# Each user can have no more than 2 active clusters
c.UserLimits.max_clusters = 2
```

If a user tries to create a cluster, and doing so would exceed a limit, the operation errors:

```python
In [6]: cluster = g.new_cluster()
...

ValueError: Cannot start new cluster, would exceed user limit of 5 active clusters.
```

If a user tries to scale a cluster to ``n` workers, and doing so would exceed a limit, the user gets the maximum number of workers, and a warning is shown:

```python
In [10]: cluster.scale(2)
/Users/jcrist/Code/dask-gateway/dask-gateway/dask_gateway/client.py:576: UserLimitWarning: Adding 2 workers to cluster be8844f83a2b4773921ae7db8f3f3a9e would exceed user cores limit of 2.0, adding 1 workers instead.
  warnings.warn(UserLimitWarning(msg["message"]))
```

jcrist:
Tests and configuration docs added, merging.

PR #102, opened 2019-09-18
Use dask-sphinx-theme

jcrist:
Use Dask's sphinx theme.

Also some content changes to make things look better with the different theme.


mrocklin:
Hooray for consistent styling!

Issue #106, opened 2019-09-18
Point `gateway.dask.org` to our docs


jcrist:
We should add `gateway.dask.org` on cloudflare to point to our docs. @quasiben I believe you have the rights for this? The docs are hosted at https://dask.github.io/dask-gateway/.

TomAugspurger:
dask-ml does something similar. We needed

1. To set `ml.dask.org` in the "Custom domain" field in the GitHub settings for the repository
2. Add a custom CNAME file to the gh-pages branch: https://github.com/dask/dask-ml/blob/gh-pages/CNAME
3. Something on the cloudflare side (I don't recall the details here. Hopefully you have access to see how it's configured).

quasiben:
Yes, add a PR now

quasiben:
cloudflare is all setup -- when #108 is merged I believe things should be functional

quasiben:
confirmed https://gateway.dask.org is up and running -- i think only @jcrist can close

jcrist:
Hmmm, I'm seeing the old docs there, not master.

- https://dask.org/dask-gateway/
- https://gateway.dask.org/

jcrist:
Got this report from github:

 The CNAME `gateway.dask.org` is already taken. Check out https://help.github.com/articles/troubleshooting-custom-domains/#cname-already-taken for more information.

jcrist:
I think this is because @quasiben pushed the cname to his own gh-pages branch in #108, which was outdated, and that's why it's serving there?

quasiben:
that's annoying -- sorry, delete the branch now

jcrist:
Hmmm, still no good. Do you have the right link on cloudflare? Getting a 404 from github now for https://gateway.dask.org/

quasiben:
I've deleted the branch and https://gateway.dask.org is getting a 404 for me.  I suspect this is a DNS caching issue now, maybe waiting a bit should resolve things

quasiben:
I checked just now and everything looks correct as best as I can tell

jcrist:
I had to force the docs to rebuild and that fixed things. Thanks @quasiben! Closing.

PR #107, opened 2019-09-18
Move to dask org

### LINKED ISSUE #105: Finish move to dask org
- [ ] Update links
- [ ] Move CI
- [ ] Move docs

jcrist:
Fixes #105.


mrocklin:
Woot!

What are some other things we should do here?  Should we add a link under the dask setup docs?

Push things to https://gateway.dask.org ?

jcrist:
> Push things to https://gateway.dask.org ?

https://github.com/dask/dask-gateway/issues/106

jcrist:
Would you want something in the dropdown? Or just in the setup docs page?

mrocklin:
Ah, I was coming back here to apologize for not seeing that yet.  Going through e-mail oldest to newest is dangerous :)

mrocklin:
I was thinking here: 

Screen Shot 2019-09-18 at 2 39 04 PM

We could even just link out directly to gateway.dask.org from there.

I think that the dropdown is also a good idea, but I would probably add that in later after dask-gateway gets more traction.  I'm happy either way though.

jcrist:
Sure, I can write something up.

jcrist:
One downside of the move is that builds are *significantly* slower, since we're now competing with dask's travis ci queue :/. May want to move to azure or something else with more builds per org.

mrocklin:
Making it easier to move projects to Azure sounds good to me. Marius was
looking at this in https://github.com/dask/community/issues/1

PR #125, opened 2019-09-23
Add client-side cluster options defaults

jcrist:
Previously all cluster options (e.g. ``image``, ``worker_cores``,
etc...) had server side defaults only. This was useful for admins to
configure what a default cluster looks like, and in many cases should be
sufficient. In some cases though a default (no kwargs) cluster may rely
on some client-side information. For example, the Pangeo deployment
would like to use the current docker image by default for all workers -
this information is only available on the client side.

This PR adds a ``gateway.cluster.options`` configuration field where
such configurations could be set. The values can be template strings,
formatted with any local environment variables. By default, all calls to
``submit``/``new_cluster`` will use the local default kwargs (with any
explicit kwargs as overrides). Calling ``cluster_options`` will get the
server side defaults, then update them with any local defaults unless
``disable_local_defaults=True`` is specified.

For the pangeo use case, a config file might look like:

```yaml
gateway:
  cluster:
    options:
      image: "JUPYTER_IMAGE_SPEC"
```

Which would then allow:

```python
```

to use the current image without further specification.

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -411,25 +411,24 @@ class Gateway(object):
         opts = dask.config.get("gateway.cluster.options")
         return {k: format_template(v) for k, v in opts.items()}
 
-    async def _cluster_options(self, disable_local_defaults=False):
+    async def _cluster_options(self, use_local_defaults=True):
         url = "%s/gateway/api/clusters/options" % self.address
         req = HTTPRequest(url=url, method="GET")
         resp = await self._fetch(req)
         data = json.loads(resp.body)
         options = Options._from_spec(data["cluster_options"])
-        if not disable_local_defaults:
+        if use_local_defaults:
             options.update(self._config_cluster_options())
         return options
 
-    def cluster_options(self, disable_local_defaults=False, **kwargs):
+    def cluster_options(self, use_local_defaults=True, **kwargs):
         """Get the available cluster configuration options.
 
         Parameters
         ----------
-        disable_local_defaults : bool, optional
-            Whether to disable using any default options from the local
-            configuration and get only the server-side defaults instead.
-            Default is False.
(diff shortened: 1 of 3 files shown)


dask-gateway/dask_gateway/client.py
```diff
@@ -407,22 +407,40 @@ def list_clusters(self, status=None, **kwargs):
...
+            options.update(self._config_cluster_options())
+        return options
 
-    def cluster_options(self, **kwargs):
+    def cluster_options(self, disable_local_defaults=False, **kwargs):
         """Get the available cluster configuration options.
 
+        Parameters
+        ----------
+        disable_local_defaults : bool, optional
```
jcrist: I'm not 100% happy with this kwarg name, if anyone has a better idea (cc @ericdill, @mrocklin). Other ideas:
- `use_local_defaults`
- `no_local_defaults`

mrocklin: At one point Aaron Meurer recommended to me that all options should be positively named, so as to avoid confusing double-negatives.  That policy would encourage the name `allow_local_defaults` instead.

mrocklin:
Is it possible for dask-gateway to allow some user options to be configurable, while disallowing some others?  If not, can we make space for this in the future?

jcrist:
> Is it possible for dask-gateway to allow some user options to be configurable, while disallowing some others? If not, can we make space for this in the future?

Yes, this is how it currently works. By default no options are configurable, the admin has to explicitly enable any parameters they want the user to be able to configure. There's also a configurable handler, so they can add whatever value checking the want (e.g. bounds, etc...).

dask-gateway/dask_gateway/client.py
jcrist: Renamed to `use_local_defaults` instead.

PR #129, opened 2019-09-27
Use `curl` backend for `httpclient` when available

### LINKED ISSUE #103: Using `CurlAsyncHTTPClient` for proxy support
At large enterprises / institutions it's often common to have proxy stuff configured via environment variables -- out of the box this causes trouble with the gateway client as it doesn't pay attention to those.

However, telling it to use `CurlAsyncHTTPClient` seems to fix the issue, by adding 

```python
AsyncHTTPClient.configure("tornado.curl_httpclient.CurlAsyncHTTPClient")
``` 

after the initial import in https://github.com/jcrist/dask-gateway/blob/master/dask-gateway/dask_gateway/client.py

Any concerns with making this the default?  According to https://www.tornadoweb.org/en/stable/httpclient.html#module-tornado.httpclient it will require newish version of `libcurl` and `pycurl`

"Currently the minimum supported version of libcurl is 7.22.0, and the minimum version of pycurl is 7.18.2."

gforsyth:
Users with local enterprise proxies often need to include additional
configuration items when passing requests. The `CurlAsyncHTTPClient`
allows for specifying a local proxy and port to tunnel requests through,
but also requires that `pycurl` is installed.

This adds a configuration section to `gateway.yaml` with some commonly
specified options and sets `CurlAsyncHTTPClient` as the default backend
if `pycurl` is available.

If someone specifies proxy configuration but `pycurl` isn't present, it
will raise an error to complain about the missing dependency, otherwise
it just uses the default tornado backend.

Suggest that with this change that `pycurl` is added to the
`dask-gateway` feedstock runtime requirements

Fixes #103

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -9,8 +9,7 @@ import traceback
 import warnings
 import weakref
 import importlib
-from datetime import timedelta, datetime
-from threading import get_ident
+from datetime import datetime
 from urllib.parse import urlparse
 
 import dask
@@ -31,11 +30,15 @@ from .utils import format_template, cancel_task
 
 del comm
 
-# Configure AsyncClient backend
+# Configure default AsyncClient backend
+HTTP_CLIENTS = {
+    "curl": "tornado.curl_httpclient.CurlAsyncHTTPClient",
+    "simple": "tornado.simple_httpclient.SimpleAsyncHTTPClient",
+}
+
+HAS_PYCURL = False
 if importlib.util.find_spec("pycurl") is not None:
-    _http_client = "tornado.curl_httpclient.CurlAsyncHTTPClient"
-else:
-    _http_client = "tornado.simple_httpclient.SimpleAsyncHTTPClient"
+    HAS_PYCURL = True
 
 
 __all__ = ("Gateway", "GatewayCluster", "GatewayClusterError", "GatewayServerError")
@@ -264,7 +267,8 @@ class Gateway(object):
         auth=None,
         asynchronous=False,
         loop=None,
-        proxy_config_dict=None,
+        http_client=None,
+        client_defaults=None,
     ):
         if address is None:
             address = format_template(dask.config.get("gateway.address"))
@@ -288,21 +292,25 @@ class Gateway(object):
             proxy_netloc = parsed.netloc if parsed.netloc else proxy_address
         proxy_address = "gateway://%s" % proxy_netloc
 
-        if proxy_config_dict is None:
-            proxy_config_dict = dask.config.get("gateway.proxy_config", {})
-        if not any(proxy_config_dict.values()):
-            AsyncHTTPClient.configure(_http_client, defaults={})
-        elif isinstance(proxy_config_dict, dict):
-            if _http_client == "tornado.simple_httpclient.SimpleAsyncHTTPClient":
-                raise ValueError(
-                    "Custom local proxy setup requires `pycurl` but it is not installed"
-                )
-            else:
-                AsyncHTTPClient.configure(_http_client, defaults=proxy_config_dict)
-        else:
-            raise ValueError(
-                "`proxy_config_dict` was specified but is not a valid dict"
+        if http_client is None:
+            http_client = dask.config.get("gateway.httpclient.type", None)
+        if http_client is None:
+            _http_client = (
+                HTTP_CLIENTS["curl"] if HAS_PYCURL else HTTP_CLIENTS["simple"]
             )
+        elif http_client == "curl" and not HAS_PYCURL:
+            raise ValueError("`curl` client requires `pycurl` but it is not installed")
+        else:
+            _http_client = HTTP_CLIENTS[http_client]
+
+        if client_defaults is None:
+            client_defaults = dask.config.get("gateway.httpclient.defaults", {})
+        elif not isinstance(client_defaults, dict):
+            raise TypeError(
+                "Expected dict for `client_defaults`, got %s" % type(client_defaults)
+            )
+
+        AsyncHTTPClient.configure(_http_client, defaults=client_defaults)
 
         self.address = address
         self.proxy_address = proxy_address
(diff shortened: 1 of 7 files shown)


gforsyth:
"But the tests pass locally..."
I'm missing something with optionally having `pycurl` available.  I'll track it down.

gforsyth:
Ok, CI is passing now.
@jcrist -- it seems like `pycurl` is a huge pain to install via `pip` and very easy to install via `conda`.  Ideally we want the `pycurl` backend to also be tested in CI, not sure if it's worth the pain of doing it via `pip` or if you are ok with a miniconda bootstrap install.

dask-gateway/dask_gateway/gateway.yaml
```diff
@@ -19,3 +19,9 @@ gateway:
   cluster:
     options: {}         # Default options to use when calling ``new_cluster`` or
                         # ``cluster_options``. Values may be template strings.
+                        # authentication class above.
+
+  proxy_config:         # Values for a config dictionary for CurlAsyncHTTPClient
+    proxy_host: null    # Local proxy host
+    proxy_port: null    # Local proxy port
+    ca_certs: null      # Custom certificates to use (available from `certifi.where()`)
```
jcrist: Since both http clients can accept many parameters, perhaps we should instead have the following fields:

```
httpclient:
  type: null         # The kind of http client to use. One of {"curl", "simple"}. If not provided, curl is used if pycurl is installed, otherwise simple is used.
  defaults: null     # A mapping of default parameters to use on every request
```

This allows supporting any of the possible parameters, without requiring them to be exposed directly in the config file.

dask-gateway/dask_gateway/gateway.yaml
```diff
@@ -19,3 +19,9 @@ gateway:
   cluster:
     options: {}         # Default options to use when calling ``new_cluster`` or
                         # ``cluster_options``. Values may be template strings.
+                        # authentication class above.
```
jcrist: I think this was copy-paste error?

gforsyth: Yep, my bad.

dask-gateway/dask_gateway/gateway.yaml
gforsyth: Sounds good.  I'll put that together.

gforsyth:
Well, on a positive note, it's much easier to `pip install pycurl` on Linux than on Mac, so we don't need to switch over to `miniconda`.  
However, switching the default backend appears to have borked some other tests.  I'll try to dig in later today.

dask-gateway/dask_gateway/client.py
```diff
@@ -29,6 +30,17 @@
...
 
+# Configure default AsyncClient backend
+HTTP_CLIENTS = {
+    "curl": "tornado.curl_httpclient.CurlAsyncHTTPClient",
+    "simple": "tornado.simple_httpclient.SimpleAsyncHTTPClient",
+}
+
+HAS_PYCURL = False
+if importlib.util.find_spec("pycurl") is not None:
+    HAS_PYCURL = True
```
jcrist: I'd avoid doing this at import time, and instead try to import when setting up the client (and erroring appropriately if not installed). This allows users to install after starting python if they forgot to, which is something I often do for other packages.

dask-gateway/dask_gateway/client.py
```diff
@@ -243,10 +255,23 @@ class Gateway(object):
...
     def __init__(
-        self, address=None, proxy_address=None, auth=None, asynchronous=False, loop=None
+        self,
+        address=None,
+        proxy_address=None,
+        auth=None,
+        asynchronous=False,
+        loop=None,
+        http_client=None,
+        client_defaults=None,
```
jcrist: Hmmm, I hadn't thought about kwarg names for these. These read a bit off to me. Perhaps:

```
httpclient_type=None,
httpclient_defaults=None
```

to better match the config names?


As written `http_client` looks like it expects an http client object, and `client_defaults` makes me ask "which client".

dask-gateway/dask_gateway/gateway.yaml
```diff
@@ -19,3 +19,7 @@ gateway:
   cluster:
     options: {}         # Default options to use when calling ``new_cluster`` or
                         # ``cluster_options``. Values may be template strings.
+
+  httpclient:
+    type: null         # The kind of http client to use. One of {"curl", "simple"}. If not provided, curl is used if pycurl is installed, otherwise simple is used.
+    defaults: null     # A mapping of default parameters to use on every request
```
jcrist: Perhaps add a link to the corresponding tornado docs here? I think http://www.tornadoweb.org/en/stable/httpclient.html#http-client-interfaces would be the right one.

jcrist: Also, please line wrap to ~80 characters

yuvipanda:
Am curious if you have considered using aiohttp rather than the built in tornado client? I've been doing that for all my projects, and a lot of tools in the ecosystem (such as kubernetes_asyncio) are too. Avoiding having to deal with curl is definitely a big reason for it...

gforsyth:
@jcrist -- I can push up changes probably tomorrow, running around a bit today.  If this is blocking a release, let me know, but I don't think it's critical.

@yuvipanda - I hadn't considered it but I can see a distinct benefit to avoiding the `pycurl` dependency.

gforsyth:
@jcrist -- pushed up the changes you requested.  I think switching to the `pycurl` backend is causing other tests to fail but to be honest I'm not sure why.

jcrist:
I can take a look later, thanks for pushing on this.

dask-gateway/dask_gateway/utils.py
```diff
@@ -14,3 +15,46 @@ def format_template(x):
...
     except asyncio.CancelledError:
         pass
+
+
+def _has_pycurl():
+    HAS_PYCURL = False
+    if importlib.util.find_spec("pycurl") is not None:
+        HAS_PYCURL = True
+
+    return HAS_PYCURL
```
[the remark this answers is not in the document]
gforsyth: just habit, but I'm happy to make that change.

gforsyth:
Hey @jcrist -- I'm going to close this for now.  I'll re-open if/when I have time to bang on it.  Commits will be on my fork if anyone wants to pick this up.

PR #146, opened 2019-10-07
[WIP] initial chartpress config

### LINKED ISSUE #101: Publish helm chart repository
A [Helm Chart Repository](https://helm.sh/docs/chart_repository/) is a static site with a simple spec. We should be able to generate one as part of the docs and host it on github pages. Built chart artifacts would then be hosted in github releases. This is similar to what [JupyterHub does](https://github.com/jupyterhub/helm-chart), except we'd host them as releases and in the main repository.

This isn't strictly necessary to use the existing helm chart, but can be nicer as it avoids users having to download the chart locally.

### LINKED ISSUE #117: Automate chart building with chartpress
In #116 we added a setup using chartpress for publishing helm charts to https://dask.org/dask-gateway-helm-repo/. It would be nice to automate this using CI.

The chart naming scheme is:

`dask-gateway-{dask_gateway_server.__version__}-{git tag}`

with the tag removed for releases (using the `--reset` flag).

jhamman:
closes #117

@jcrist - this is a start at setting up chartpress. I think you'll need to setup the encryption keys and any dns settings. I'll leave some inline comments below.


continuous_integration/kubernetes/deploy.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+echo "Deploying chartpress..."
+
+set -eu
+# TODO: add deploy-key.rsa.enc and set keys in travis UI
+openssl aes-256-cbc -K $encrypted_058f1ad78f7f_key -iv $encrypted_058f1ad78f7f_iv -in deploy-key.rsa.enc -out deploy-key.rsa -d
```
jhamman: @jcrist - this is the main thing you'll need to do.

jcrist: Any reference on what I need to do here to make this work?

jcrist:
One question I have is when should we be building dev images/chart releases, and what should they be tagged as? Building and uploading a new docker image for every merged commit of `dask-gateway`/dask-gateway-server` seems unnecessary, but some changes to the helm chart may require the master branch of `dask-gateway-server` to work properly.

A few options:
- Build new images for every merged commit, but always write to the `dev` tag. Docker-Hub will only have releases and the master branch available, but this should prevent us from using too much of docker-hub's resources.
- Never build a dev image unless explicitly asked to. We could do this with a commit-message tag (as we do for other tests). Some thing like `[build-docker]`.
- Only build a dev image when we build a new helm-chart version. Images and dev-releases of the helm chart share the same tag, and will persist indefinitely on docker-hub. We'd probably want a way to trigger a helm chart release even if the chart itself didn't change (say there's some patch to dask-gateway-server that we want people to be able to try out). This is a bit tricky since it ties the helm chart to the master branch, which may not be ready for dev use.

Right now I'm thinking either option 1 or option 3.

continuous_integration/kubernetes/deploy.sh
jhamman: @jacobtomlinson setup our pangeo chart: https://github.com/pangeo-data/helm-chart/pull/1

In travis CI, we set 4 env variables:

image

jacobtomlinson:
I think option one sounds most reasonable. It's good to have a dev image available, but you don't need them all forever.

continuous_integration/kubernetes/deploy.sh
TomAugspurger: @jcrist do you mind if I do this? I don't think it matters who (if anyone) has the unencrypted deploy key.

jcrist: I was just looking into this. I'll take care of it, I want to make sure it integrates nicely with our CI setup (which is a bit complicated).

continuous_integration/kubernetes/install.sh
```diff
@@ -25,5 +25,9 @@ if [[ "$TRAVIS" != "true" ]]; then
     eval $(minikube docker-env)
 fi
 pushd $git_root
+
+echo "Installing chartpress"
+pip install --default-timeout=100 git+https://github.com/jupyterhub/chartpress.git
```
jacobtomlinson: This can just be `chartpress` now as it is on PyPI.

```suggestion
pip install --default-timeout=100 chartpress
```

continuous_integration/kubernetes/script.sh
```diff
@@ -2,3 +2,14 @@
 
 echo "Running tests..."
 kubectl exec dask-gateway-tests -n dask-gateway /working/continuous_integration/kubernetes/script-internal.sh
+
+echo "Running chartpress"
+pushd $TRAVIS_BUILD_DIR/resources/helm/
+docker build -t dask-gateway-base images/dask-gateway-base
+chartpress --commit-range ${TRAVIS_COMMIT_RANGE} --tag `date +%y.%m.%d`
```
jacobtomlinson: Did you intend to use CalVer here? I would expect we want to use the git tag instead.

```suggestion
chartpress --commit-range ${TRAVIS_COMMIT_RANGE} --tag ${TRAVIS_TAG}
```

resources/helm/chartpress.yaml
```diff
@@ -1,5 +1,13 @@
 charts:
   - name: dask-gateway
+    imagePrefix: daskgateway/
     repo:
       git: dask/dask-gateway-helm-repo
       published: https://dask.github.io/dask-gateway-helm-repo
```
jacobtomlinson: I raised dask/dask-gateway-helm-repo#1 to say that it would be great to deploy the charts over to dask/helm-chart instead so that all Dask related charts are served from https://helm.dask.org.

It would be great to update these values to point there instead. This will mean that the deploy key needs to be for that repo instead though.

```
      git: dask/helm-chart
      published: https://helm.dask.org
```

continuous_integration/kubernetes/deploy.sh
jacobtomlinson: You'll want to generate an ssh key for Travis to use to deploy.

```bash
ssh-keygen -t rsa -b 4096 -q -N "" -C "" -f continuous_integration/kubernetes/deploy-key.rsa
```

You'll need to create a deploy key over on https://github.com/dask/helm-chart with the contents of the `deploy-key.rsa.pub` file that gets generated.

Then use the Travis CI CLI tool to encrypt it. It will use openssl to encrypt it and automatically add the decryption keys to Travis CI as environment variables.

```bash
travis encrypt-file continuous_integration/kubernetes/deploy-key.rsa --com --add

# You might need to move the encrypted file to the correct location as it will probably
# just drop it in your working directory
mv deploy-key.rsa.enc continuous_integration/kubernetes/deploy-key.rsa.enc
```

You may also want to add the `deploy-key.rsa` and `deploy-key.rsa.pub` files to the `.gitignore` for safety.

Lastly the Travis CI CLI command will print out the openssl line that you need to replace this one with here. Again the path to the `deploy-key.rsa.enc` might be wrong and need changing to `continuous_integration/kubernetes/deploy-key.rsa.enc`.

jcrist:
Apologies for never getting around to this. Since this is a bit stale and conversations about automating releases are scattered, I'm going to close this in favor of consolidating conversation in #296. We might want to reopen/use these commits later, but closing for now.

PR #154, opened 2019-10-14
Add preemptible worker example

TomAugspurger:
cc @jcrist, @jacobtomlinson @jhamman .

A few questions

1. Scope: I included a small example creating a preemptible node pool with gcloud. Do we want that, or is it out of scope for our docs? Typically we point to zero-to-jupyterhub, but they don't discuss preemptible nodes
2. New config: Can the experts (@jacobtomlinson, @jhamman 😄) check that the config makes sense. I used this on a demo project and things worked well I think. I'm not sure if everything is necessary though. In particular, I'm not sure if the config for the scheduler is needed / correct.

### DIFF
--- a/docs/source/install-kube.rst
+++ b/docs/source/install-kube.rst
@@ -279,6 +279,61 @@ object.
     ... )
 
 
+Using Preemptible Nodes for Workers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some cloud providers offer "preemptible" VMs. These are VMs with limited lifetimes that may
+be terminated with little warning, but cost much less than standard VMs. This tradeoff can be
+well-suited for Dask Workers (but not the scheduler, or for dask-gateway itself). To use preemptible
+resources for your workers, use Kubernetes' `taints and tolerations`_.
+
+Your node pool will need the taint. For example, with the ``gcloud`` CLI, this is
+
+.. code-block::
+
+   gcloud beta container node-pools create  \
+       ... # other create options
+       --preemptible \
+       --enable-autoscaling --min-node=<min-nodes> --max-nodes=<max-nodes> \
+       --node-taints preemptible=true:NoSchedule
+
+Then in your ``config.yaml`` for dask-gateway's Helm chart, you'll need to specify the following to
+
+1. Ensure that the scheduler *does not* end up on a preemptible node.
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
At a glance that all looks reasonable to me. 

One note is that Preemptible Nodes is a Google-specific term. Azure call them low-priority VMs and AWS call them Spot Instances. I wonder if there is a more generic collective term for them.

rsignell-usgs:
@TomAugspurger and @jacobtomlinson , I'm guessing this one slipped of the radar screen?

TomAugspurger:
Yeah... just dropped it. Are you / qhub looking to deploy something with it?

I think for the GCP deployment, we're using https://github.com/pangeo-data/pangeo-cloud-federation/blob/ac490dabf56caa8fd1ef6dfbcd191d974a407c85/deployments/gcp-uscentral1b/config/common.yaml#L166-L175 and GKE's "autoprovisioning" (https://github.com/pangeo-data/pangeo-cloud-federation/blob/ac490dabf56caa8fd1ef6dfbcd191d974a407c85/deployments/gcp-uscentral1b/Makefile#L46) which magically creates node pools to match node tolerations.

The terraform I shared for AKS at https://gist.github.com/TomAugspurger/3a77bb89f75f07667107f83cad534e80 has a spot pool.

rsignell-usgs:
@TomAugspurger , cool, this came up yesterday in some discussions with USACE folks, who have been working on Azure, so I'll point them this way.

dgerlanc:
> At a glance that all looks reasonable to me.
> 
> One note is that Preemptible Nodes is a Google-specific term. Azure call them low-priority VMs and AWS call them Spot Instances. I wonder if there is a more generic collective term for them.

Preemptible is probably the most accurate term for how these instances behave, even though it's Google specific. Or "interruptable" maybe.

consideRatio:
@TomAugspurger I'm triaging a bit. Is this PR something you want to champion still?

I'm a bit hesitant on documentation this detailed as I've found it a bit too hard to maintain over time when it includes details about specific cloud providers etc that may change. I'm fine with merging if its updated to resolve merge conflicts and the gcloud commands etc still works as documented. If it remains stale I suggest we close this though.

TomAugspurger:
I’m fine with closing!

consideRatio:
Thanks for followup @TomAugspurger!

PR #164, opened 2019-11-21
Allow dashboard and REST addresses to differ

jcrist:
This adds configuration to distinguish between the address the python
client should interact at from the address used when formatting
user-facing links.

This can be useful when running in an environment where the client
process should access the dask-gateway server at a different address
than the user's browser (e.g. in a JupyterHub setup, where Python
kernels are remote).

For example, if dask-gateway should be accessed behind JupyterHub's
proxy when outside the cluster, the user-side configuration would look
like:

```yaml
gateway:
  public-address: /services/dask-gateway/
  # ...
```

For now this is a configuration-only field, as I don't expect it to be
used by most users.

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -202,7 +202,7 @@ class ClusterReport(object):
         )
 
     @classmethod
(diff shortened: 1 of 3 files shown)


jcrist:
cc @martindurant.

PR #182, opened 2019-12-04
Use k3d instead of minikube for testing

jcrist:
Trying out using [k3d](https://github.com/rancher/k3d) instead of minikube for testing. I've found minikube to be flaky and slow, hoping this helps make the development experience on kubernetes a bit nicer.


jcrist:
This drops around 2 mins off our CI time for kubernetes runs. If you're developing locally and already have k3d and its images around, setting up a CI cluster happens in a couple seconds rather than minutes, and generally seems less prone to random failures than minikube.

I can see a few downsides though:
- No exposing of services through load balancers (only nodeports, ports must be known prior). Ingress does work though.
- The CLI seems a bit unstable, the upcoming release changes a few things. This is fine for a new tool (our CI script can just pin a version), but it's not as stable as minikube.
- Devs may be more likely to have minikube already?
- Uses a recent kubernetes version (1.16). I'm not sure what versions are commonly deployed and if we should test on older versions (we currently test on 1.14).

I could split the k3d/minikube parts out so we could support both if people have strong preferences on dev tools, but would prefer not to if possible.

cc @jacobtomlinson, @yuvipanda for kubernetes knowledge.

jacobtomlinson:
I'm a big fan of k3s/k3d so definitely support this change 🎉.

> No exposing of services through load balancers

This is a bit of a shame and I can foresee usecases where this will be frustrating.

> Devs may be more likely to have minikube already

I get the impression that the built in k8s support in Docker for Mac/Windows has overtaken minikube already. Ideally, folks should be able to develop against any k8s implementation and get a consistent experience. So hopefully over time this will matter less.

> Uses a recent kubernetes version (1.16)

I assume they just didn't bother backporting older versions. I assume versions going forwards will stick around, so this will not be an issue in the future.

> I could split the k3d/minikube parts out so we could support both if people have strong preferences on dev tools, but would prefer not to if possible.

I'm not sure if this is reasonable today, but I would like to not care about what k8s implementation people are using. The CI can use one tool, local devs can use another, deployments will use a whole range of things. It would be nice to support k8s versions and note care past that (perhaps with some minimum rules like load balancers should become routable and the default volume driver must attach volumes).

jcrist:
Thanks @jacobtomlinson, it's good to hear your perspective.

We don't need anything special here kubernetes-wise, so people can already develop against any of the options you mentioned. For now I'm going to merge this. The scripts are mostly for CI, so our choice here shouldn't affect the dev experience too much. Later on it might be nice to clean these up for local usage.

PR #195, opened 2020-02-04
Rearchitecture

### LINKED ISSUE #186: Rearchitecture
After a few months working on this, a few past design decisions are starting to cause some issues.
- #148: we're currently limited to a single server model, with the same restrictions as JupyterHub (this is because much of the design is based on JupyterHub).
- Even though we *can* support a database for persistence, setting up additional infrastructure to do so places an additional burden on administrators. People are unlikely to do so unless necessary.

At the same time, the external facing API seems fairly solid and useful. I've been trying to find a way to rearchitect to allow (but not mandate) running multiple services, without a (mandatory) database to synchronize with, and I think I've reached a decent design.

## Design Requirements

- We must be able to support all major backends using the same codebase (as much as possible)
  - Kubernetes
  - Hadoop
  - HPC JobQueues
  - Local processes
  - Extension points for custom backends
- We must keep install instructions simple. A basic local install must not require installing anything that can't be installed from PyPI/conda-forge.
- Performance in a single server-instance must not be overly degraded by any changes. Currently we've optimized for the single-server model, so I expect to lose some performance here, but it would be good to keep this use case in mind.
- Where possible additional infrastructure (e.g. a database) should not be required. If we can rely on existing systems found for each backend we should.
- When properly configured, each backend should have a way for the gateway to run without a single point of failure (at least one written by us). I don't see HA for *most* deployments increasing scalability, but I do see it improving robustness.

Below I present a design outline of how I hope to achieve these goals for the various backends.

## Authentication

As with JupyterHub, we have an `Authenticator` class that configures authentication for the gateway server. If a request comes in without a cookie/token, the authenticator is used to authenticate the user, then a cookie is added to the response so the authenticator won't be called again. The decrypted cookie is a uuid that maps back to a user row in our backing database.

I propose keeping the authenticator class much the same, but instead of using a uuid as a cookie, I propose using a [jwt](https://jwt.io/) to store the user information. If a request contains a jwt, it's validated, and (if valid) information like user name and groups can be extracted from the token. This removes the requirement for a shared database to map cookies to user names - subsequent requests will already contain this information.

## Cluster IDs

Currently we mint a new `uuid` for every created cluster. This is nice as all backends have the same cluster naming scheme, but means we need to rely on a database to map uuids to cluster backend information (e.g. pod name, etc...).

To remove the need for a database, I propose to encode backend information in the cluster id. This means that each backend will have a different looking id, but means we can parse the cluster id to reference it in the backing resource manager instead of using a database to map between our id and the backend's.

For the following backends, things might look like:
- Kubernetes: Cluster CRD object name
- Hadoop: application id
- HPC job queue/local processes/etc...: requires a database, probably same id scheme as now.

## Cluster Managers

To support multiple dask-gateway servers, I found it helpful to split our backends into two categories:

**Useful database in the resource manager**
- Kubernetes
- YARN

**No useful database in the resource manager**
- Local processes
- HPC Job queues

The former category could support all our needed functionality without any need for synchronization outside of requests to the resource manager. The latter would require additional infrastructure on our end if we wanted to run multiple instances of the gateway server.

Walking through my proposed ideal implementations for each backend:

### Kubernetes

The proposed design for running dask-gateway on kubernetes in an *ideal* (IMO) deployment is as follows:

- A dask-gateway deployment, containing one or more instances of our gateway server.
- [Traefik](https://containo.us/traefik/) proxy deployment, containing one or more pods running traefik, configured with an [IngressRoute](https://docs.traefik.io/providers/kubernetes-crd/) provider. Traefik 2.0 added support for proxying TCP connections dispatched on SNIs, which means it can support all the things we needed for our own custom proxy implementation. The old proxy could still be used, but doesn't (currently) support multiple instances. Traefik handles all this for us in a scalable manner.
- A CRD for a dask cluster, and a backing controller (probably written with [metacontroller](https://metacontroller.app/)). Making dask clusters a resource allows us to better use kubernetes to store all our application information, removing the need for an external database. It also removes the need to synchronize our application state and the resource manager's state - they're identical in this case. Operations like scaling a cluster are just a patch on the CRD. The metacontroller web hooks could run in the `dask-gateway-server` pods, or could be their own pods - right now I'm leaning towards the former for simplicity.

Listing clusters, adding new clusters, removing old clusters, etc... can all be done with single kubernetes api requests, removing a lot of need for synchronization on our end. It also means admins can use kubectl without worrying about messing with a deployment's internal state.

### Hadoop

The Hadoop resource manager could also be used to track all the application state we care about, but I'm not sure if querying it will be performant enough for us. We likely want to support an optional (or maybe mandatory) database here. Some benchmarking before implementation will need to be done.

In this case, an installation would contain:
- One or more dask-gateway server instances, behind a load balancer (load balancer not needed if only running one instance).
- Our own existing custom proxy, or traefik. If using traefik we could maybe make use of JupyterHub's work to make it [configurable via a file provider](https://github.com/jupyterhub/traefik-proxy).
- Management of individual clusters will be moved to [skein's application master](https://jcrist.github.io/skein/specification.html#specification-master), instead of the gateway server. This will work similar to the kubernetes version above - scaling a cluster will forward the request the application master to handle. Querying for the current scale will query the application master. This will likely increase latency (all requests will have to ping an external service), but remove the bottleneck on the gateway server for handling everything.

### HPC job queue systems

HPC job queue systems will require an external DB to synchronize multiple dask-gateway servers when running in HA mode. With some small tweaks, this will mostly look like the existing implementation. I plan to rely on Postgres for this, making use of its `SKIP LOCKED` feature to implement a work queue for synchronizing spawners, and [asyncpg](https://github.com/MagicStack/asyncpg) for its fast postgres integration.

## Cluster backend base class

Currently we define a base `ClusterManager` class that each backend has to implement. Each dask cluster gets its own `ClusterManager` instance, which manages starting/stopping/scaling the cluster.

With the plan described above we'll need to move the backend-specific abstractions up higher in the stack. I propose the following initial definition (will likely change as I start working on things):

```python
class ClusterBackend:
    async def list_clusters(self, user=None):
        """List all running or pending clusters.

        Parameters
        ----------
        user : str, optional
            If provided, filter on this user name.
        """
        pass

    async def get_cluster(self, user, cluster_id):
        """Get information about a cluster.

        Same output as `list_clusters`, but for one cluster"""
        pass

    async def start_cluster(self, user, **cluster_options):
        """Start a new cluster for this user"""
        pass

    async def stop_cluster(self, user, cluster_id):
        """Stop a cluster.
        
        No-op if already stopped/doesn't exist.
        """
        pass
        
    async def scale_cluster(self, user, cluster_id, n):
        """Scale a cluster to `n` workers."""
        pass
```

Things like `kubernetes` and maybe `hadoop` would implement the `ClusterBackend` class. We'd also provide an implementation that uses a database to manage the multi-user/cluster state and abstracts on a single-cluster class (probably looking the same as our existing `ClusterManager` base class).

```python
class DatabaseClusterBackendBase(ClusterBackend):
    """A backend class that uses a database to synchronize between users/clusters.

    HPC job queue backends would use this, as well as other backends lacking a
    sufficiently queryable resource manager."""
    cluster_manager_class = Type(
        klass=ClusterManager,
        help="The cluster manager to use to manage individual clusters"
    )
```

jcrist:
Fixes #186.

This rearranges our backend abstraction (as well as some of the control flow) to make it easier to make the kubernetes backend (and other backends) more native/performant.

- [x] Redo authentication to remove need for database
- [x] Abstract out backend class
- [x] Move database to be backend-specific, rather than required by the gateway app itself
- [x] Switch from tornado to aiohttp/asyncio everywhere
- [x] Local process backend
- [x] YARN backend
- [x] PBS backend
- [x] Slurm backend
~[ ] User limits~
~[ ] Kubernetes backend~
- [x] Tests
~[ ] Docs~

For the most part this is fully-functional already, and should *just work*. The only breaking changes here are on the administrator side - the user api remains the same.

I might drop the kubernetes support on master for a bit before we work things back up. The kubernetes backend will be split between the server and the a controller, I don't really want to rewrite the kubernetes backend to use the db only to remove it in a subsequent PR. I'm hesitant to continue this work on a feature branch since I don't want to deal with huge merge conflicts down the line, and this PR is already big enough. Either way tests and docs will be updated accordingly before merging.

I'll write up a design overview of what this PR brings tomorrow, it's late here.


mmccarty:
Thanks @jcrist I agree on dropping kubernetes support in master while this work is in flight rather than a long running feature branch.

jcrist:
Ok, I think this is good to merge now (provided tests pass). I've been heads-down getting this refactor done (apologies for ignoring other things for a bit). Refactoring the world is never fun, and this PR had to touch almost everything.

For users this should have no visible changes, but this does completely change the configuration/deployment story for admins. As such the docs will have to be updated. I've turned off doc builds on master for now, since we don't have the ability to publish multiple docs versions. I think when everything is up to date we can re-enable things.

A few features that existed in the previous release have been dropped in this refactor. They can be added in subsequent PRs:

- Kubernetes support (large amount of work, will file a separate issue detailing the design)
- User resource limits (need to change implementation for new design)

A summary of changes included in this PR:

## Dependencies

We switched from using `tornado` to using `aiohttp` for our web server/client library. Benchmarks show this to be a mild performance improvement on the server side, and a moderate one on the client side (note - this is http client requests in the *server* only, updating the `dask-gateway` client library to use aiohttp is a separate issue). The aiohttp client also doesn't depend on pycurl, which had been problematic to install in the past.

## Authentication

Authentication has been rewritten to remove the need for a backing database. External authenticator classes will likely need to be changed, but the changes should be minimal (function signature, switch to aiohttp handlers).

Clusters now have a `username` associated with them, but a "user" isn't a persistent object in the gateway data model. This removes any inconsistency issues between dask-gateway's user table and the upstream authentication system (e.g. JupyterHub). To reduce load on the upstream system, the server will cache authenticated users for a configurable period of time (5 min by default). Since authentication is transparent to the user, forcing periodic reauthentication shouldn't cause any issues (thanks @yuvipanda for the idea!).

User groups have also been added to the authentication model, but we're not doing anything interesting with them yet. This could be used to provide view/modify access, shared clusters, etc...

## Backends

Previously we abstracted on a `ClusterManager` interface, which provided methods for managing an individual dask cluster (each new dask cluster would get a new `ClusterManager` instance). The `DaskGateway` application contained methods for querying/managing many clusters across users.

We now extract both of these out into a new `Backend` class. A single `Backend` is created when `dask-gateway` starts up. The backend contains methods for starting/stopping/querying clusters in bulk (see https://github.com/dask/dask-gateway/pull/195/files#diff-a469bacde3a6b41a2553ba319fb23f91R127-R206). Note that it does *not* require a database, implementations are free to store information in whatever way they see fit.

We also provide an implementation of the base `Backend` interface that stores state in a database. This is used to implement the existing backends (local processes, YARN, SLURM, PBS, etc...). More on this in the next section.

Support for user-configurable cluster options has also been expanded to support per-user or per-group options.

## Database Backend

The current non-kubernetes backends all subclass from `dask_gateway_server.backends.db_base.DatabaseBackend`, which uses a configurable database to manage cluster state. This is similar to the previous implementation, but has a few changes to make things more robust/scalable.

- State transitions are managed using the reconciler pattern (as done in kubernetes controllers). Clusters and worker records each store their current `status` (e.g. `SUBMITTED`) along with a `target`  (e.g. `RUNNING`). As events trigger on the server (whether from periodic tasks or incoming requests), these records may be updated and the object (cluster or worker) queued for processing. The processing queues deduplicate queued tasks, so if many events happen for cluster X, only the final state will be processed when it's popped off the queue.
- Failed workers are now restarted, meaning each individual dask cluster is more resilient to failure.
- Interactions between the gateway server and the individual dask clusters have been rearranged to minimize outgoing requests from the gateway server to an individual dask cluster. Most communication is now initiated by the individual clusters (as heartbeats) rather than the gateway (as polling).
- Previously the gateway server could be killed and restarted without losing active clusters  - on restart it would check the status of each cluster that was running before, and update the database state accordingly. This worked fine, but could be slow for a large number of active clusters. This failover model is still supported, but is now much much quicker. A newly started gateway process is ready to respond to requests immediately, failures that happened during the downtime are discovered and cleaned up lazily.
- There is a limiting parallelism factor `c.DatabaseBackend.parallelism` that sets the number of processing workers in the gateway. This limits the number of start/stop workers/clusters that can be processing at any time. Previously we didn't limit this, requests to start a worker would queue immediately and may fail under high load. This limit may increase latency under high load (as things back up), but should reduce the cases of failure.

## Proxy

A dask-gateway instance no longer has a `Proxy` object as a top-level configurable thing. Rather, the proxy belongs to the `Backend`, and may or may not be managed by it.

For non-kubernetes backends, the `DatabaseBackend` class manages a refactored version of our custom proxy implementation. A few changes have been made to make this more robust and scalable. 

- Both the HTTP proxy and TLS proxy have been merged into a single process, rather than 2 separate processes.
- Instead of having the gateway server *push* route updates to the proxy server, the proxy server now *pulls* updates from the gateway server. When a proxy instance starts up, it pings a configured URL on the gateway server to request the routing table. Once received, it opens a streaming request to get updates to this table. All route change events are versioned, so if a connection fails it can restart from the last change it saw (this protocol borrowed and modified from how kubernetes handles watch streams). One perk of this design is that it allows for running multiple instances of the proxy transparently to the gateway server, as the gateway server just sees this as another request.

jcrist:
I don't expect anyone to be able to review a +6,608/−8,575 line PR, so I'm going to merge this for now. The master branch will be less featureful than the previous release until we re-add these features, but I'd rather work on the master branch than a feature branch.

PR #200, opened 2020-02-13
Use aiohttp for http client in `dask-gateway` client library

### LINKED ISSUE #103: Using `CurlAsyncHTTPClient` for proxy support

### LINKED ISSUE #193: "HTTPClientError: HTTP 403: Use Proxy Server" due to local HTTP_PROXY?
Dask-gateway feels like a great fit for a project I'm involved in, and I'm excited to try it.  Thank you!

I've followed the [Kubernetes + Helm instructions](https://gateway.dask.org/install-kube.html#install-the-helm-chart) to install the server out in the wild, and I'm getting a `HTTPClientError: HTTP 403: Use Proxy Server` when I [try to connect to the public gateway](https://gateway.dask.org/install-kube.html#connecting-to-the-gateway) from my local machine.  I can instantiate the `gateway` just fine, but I can't `.list_clusters()` or `.new_cluster()` without the error above.

I'm behind a corporate web proxy that I usually handle with an `HTTP_PROXY` environmental variable.  Could this be related to my error?  If so, is there a way for me to handle a local web proxy so that I can connect my local gateway client to the external public gateway?

Thanks again.

jcrist:
We swap over to using the aiohttp Client instead of tornado's http client for making requests in the `dask-gateway` client library. The aiohttp client is more performant, has more features, and is easier to install since it doesn't require pycurl.

This also lets us delete a bunch of code, since we don't need to manage our own cookiejar implementation anymore.

Fixes #103. Fixes #193.


jcrist:
This doesn't fully fix #193 or #103 (yet). We'll need to expose configuration options on the client side for forwarding things like `proxy` and `proxy_auth`. We can also infer the http proxy configuration from environment variables, although I'm not sure if this should be on or off by default.

---

However, now that I look at the aiohttp docs, I'm not sure if we want to go through with this change. Currently aiohttp only supports:

> aiohttp supports plain HTTP proxies and HTTP proxies that can be upgraded to HTTPS via the HTTP CONNECT method. aiohttp does not support proxies that must be connected to via https://.

(from https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support)

I'm not sure if this support is sufficient for most use cases, or if not having direct https proxy support is a deal breaker. cc @gforsyth for thoughts.

gforsyth:
@jcrist I talked to a few folks who know more about enterprise proxies than I do and the consensus is that this is fine.  Seems like the upgraded connection via `HTTP CONNECT` is a common pattern and so should support most use cases that need to go through a proxy from the client side.

gforsyth:
> We can also infer the http proxy configuration from environment variables, although I'm not sure if this should be on or off by default.

My expectation is that this is on by default, but so long as error messages are sufficiently informative that it isn't, then either way works.

jcrist:
Thanks @gforsyth. I've added http_proxy support to this PR. By default we infer proxy configuration from the user's environment, but this can be disabled by setting `gateway.http-client.proxy` to `False` in the user's dask configuration. That field can also be used to explicitly set the proxy configuration, skipping inference from the user's environment.

PR #206, opened 2020-02-18
Kubernetes backend rewrite

### LINKED ISSUE #198: Rearchitecture: Kubernetes Backend

jcrist:
This implements (or will) the design proposed in #198. The kubernetes backend is broken into 2 components:

- A `Backend` subclass, responsible for creating and querying `DaskCluster` objects
- A controller (written in go) that processes and manages these `DaskCluster` objects

The controller is responsible for the following:

- When a new DaskCluster object is created, it creates a backing secret containing temporary credentials
- After the secret is created, it creates a scheduler pod, following the pod template in the `DaskCluster` object. The secret is mounted as a volume named `dask-credentials` in this pod.
- When the scheduler pod is created, additional ingress routes should be added (not specified in the CRD yet).
- Upon scale-up events, instances of the worker pod are created based on the worker pod template. Like the scheduler pod, the worker pods must mount the secret as a volume named `dask-credentials`.
- The controller should restart any pod that exits with a non-zero exit code. Additionally, it should ensure that the number of worker pods is never less than `cluster.spec.worker.replicas`.
- On a stop-cluster event (triggered by a patch of the CRD, this field isn't implemented yet), all pods should be stopped, and the ingress routes and secret should be deleted.
- On a deletion event, all child objects (secret, ingress routes, pods) should also be deleted. This could be done with ownerReferences, so nothing for the controller to do.

Fixes #198.

*(originally opened as #205, closed and reopened from a branch hosted on the dask upstream repo to make it easier for other contributors. Github doesn't let you change the source branch on an open PR)*


dask-gateway-k8s-operator/pkg/apis/gateway/v1alpha1/daskcluster_types.go
```diff
@@ -0,0 +1,55 @@
...
+type Worker struct {
+	Template corev1.PodTemplate `json:"template"`
+	Replicas int32              `json:"replicas,omitempty"`
+}
+
+// DaskClusterSpec defines the desired state of DaskCluster
+type DaskClusterSpec struct {
+	Scheduler Scheduler         `json:"scheduler"`
+	Worker    Worker            `json:"worker"`
+	Info      map[string]string `json:"info"`
```
droctothorpe: I'm not sure that declaring subtypes for Scheduler and Worker adds anything. 

Something like this might make more sense:
```golang
// DaskClusterSpec defines the desired state of DaskCluster
type DaskClusterSpec struct {
	Scheduler corev1.PodTemplate `json:"scheduler"`
	Worker    corev1.PodTemplate `json:"worker"`
	Info      map[string]string  `json:"info"`
	Replicas  int32              `json:"replicas,omitempty"`
}
```

It's clear the replicas refers to workers, but we can always label it workerReplicas if necessary.

droctothorpe:
@jcrist What flags did you pass when you ran `operator-sdk add api`?

jcrist:
> @jcrist What flags did you pass when you ran operator-sdk add api?

```
operator-sdk --api-version=gateway.dask.org/v1alpha1 --kind=DaskCluster
```

droctothorpe:
There are random things missing that should be there when you run `operator-sdk new`, like the `build` directory and the `Dockerfile` in it. The lack of that file, in particular, throws off downstream commands like controller generation and `operator-sdk build`. There's also a weird bug with the generated crd where it winds up being thousands of lines long and therefore cannot be deployed. 

Did you update any files other than `daskcluster_types.go`? If not, it might make sense to regenerate the scaffolding and pull that data in.

dask-gateway-k8s-operator/pkg/apis/gateway/v1alpha1/daskcluster_types.go
droctothorpe: Hm, on second thought, the type will probably need to be a byte slice. The controller is going to have to deserialize the contents into corev1.PodTemplates.

jcrist: I originally had that, but wanted to match other kubernetes objects that used pod templates (where there's a `template: ...` field the template goes under). Since there will be other things we'll need to add to the spec (ports exposed on the scheduler, etc...), I'm inclined to leave the nesting for now unless it proves problematic.

jcrist:
Ah! Good catch. The `.gitignore` file in the repository root was ignoring the build directory for the kubernetes operator, fixed now.

> There's also a weird bug with the generated crd where it winds up being thousands of lines long and therefore cannot be deployed.

Yeah, that's what the `build_tools/reformat_crds.py` script is for. This reformats the crd files to make them shorter (better for git diffs) and smaller in bytes (by removing the description fields) so that kubectl apply works. There's a limit of 256 KiB to files that are deployed with `kubectl apply`, since in that case the contents are also written as an annotation to provide create-or-update semantics, and annotations have a limit of 256 KiB. This limit is not present for `kubectl create`. Either way, using the reformat script I wrote fixes things.

dask-gateway-k8s-operator/pkg/apis/gateway/v1alpha1/daskcluster_types.go
droctothorpe: Copy that. Makes sense!

droctothorpe:
@jcrist Looks like the verbose CRDs were a result of passing `corev1.PodTemplate` as the type for `Scheduler` and `Worker` in the CRD struct.

When I switched it to a type of `string`, the resulting CRD was only 71 lines long.

The controller will deserialize the strings using `json.Unmarshal` from the standard library. 

Do you want one, large, fully-functional PR / POC or would you prefer smaller PRs along the way?

jcrist:
> Looks like the verbose CRDs were a result of passing corev1.PodTemplate as the type for Scheduler and Worker in the CRD struct.
> When I switched it to a type of string, the resulting CRD was only 71 lines long.
> The controller will deserialize the strings using json.Unmarshal from the standard library.

Yes, this is correct. But the cleaned up CRD files work fine, and shouldn't cause any problems. Why would you want to serialize the pod specs as a string instead of the full object? This way things still deserialize properly, and we get schema verification on object creation rather than having to handle that ourselves in the controller.

Further, things other than the controller have to work with these objects - the gateway api server watches them too. Having the templates directly as objects rather than strings that need further work to interpret makes this nicer.

I guess I don't understand why you'd want to define them as strings over the full spec.

> Do you want one, large, fully-functional PR / POC or would you prefer smaller PRs along the way?

I would prefer if you formed a WIP branch and made a PR against this branch. This allows us to discuss code as you work on it, and makes it easier to incrementally merge things upstream when we find good merge points.

droctothorpe:
Copy that. It's really helpful to hear the rationale behind the decision. Thanks for the clarification!

droctothorpe:
Any particular reason that we're using a pod template as opposed to a corev1.pod?

jcrist:
Pod templates are commonly used for other objects that create many instances of a pod from a template (jobs, deployments, statefulsets, etc...). A pod is an instantiation of a pod template.

jcrist:
I have run into a problem that I'm not sure how to solve.

To connect to a cluster, a user needs the temporary credentials we generate for it (TLS key and cert). These need to be stored in kubernetes, and also accessed by the gateway api server. Because we want the gateway to be able to manage clusters in multiple namespaces, this presents a bit of a problem.

For a pod to access the secret, it must be in its own namespace. Thus, if we only store the TLS credentials in a secret, then the api server needs a cluster role to read (`list`, `get`, `watch`) secrets for *all* namespaces. This is a huge level of access, and not something that a security-minded admin might like.

One alternative option is to store the credentials inside the `DaskCluster` custom resource object. This means that any user that has permissions to read `DaskCluster` objects in the namespace the gateway is deployed in can access anyone's cluster. However, if you have that access you probably are already an admin anyway. Since they're not stored in standard objects (pods, configmaps, ...) it seems unlikely that other users would be given access to read these objects. This allows us to give the api server only read/modify credentials for `DaskCluster` objects in its own namespace, which simplifies our security access across the cluster as well as simplifies our code (only need to watch one resource in one namespace).

Is this a bad idea? Keeping a secret in something common like a configmap is a bad idea since lots of users need access to configmaps. But storing a secret in a custom resource object that is only needed for the application (and other users are unlikely to have access) seems less flawed. AFAICT secrets are just like any other object in kubernetes, except that it's a convention to store secret info in them.

cc @yuvipanda, @jacobtomlinson for thoughts.

jcrist:
A third option would be to create the secrets in the same namespace as the DaskCluster objects, and have the controller mirror them to namespace the cluster eventually runs in. This is more complicated than storing them in the DaskCluster objects, but keeps secret things stored in secrets. This also allows for admins who can query/modify DaskCluster objects but can't connect to them, since the permissions are separated (could be given `daskclusters` permissions but not `secrets`).

I guess I'm looking for feedback from someone who knows more about what kubernetes admins might expect. Storing everything in the DaskCluster object makes our code the simplest, so if that's fine then I'd prefer that.

droctothorpe:
I lean towards storing some credentials in the CR. I don't think that pattern is unheard of and I'm not sure there's a need to create separate service accounts for access to the CR vs the TLS key and cert. Take my input with a grain of salt though. I need to do some more digging. I'll also sync up with our SRE team to see if they have any objections.

jcrist:
> I'll also sync up with our SRE team to see if they have any objections.

That would be great, those are the kind of people I want to hear from.

droctothorpe:
Fourth option: create another CRD (when you're a hammer, everything looks like a nail) that shares the same namespace as the `DaskCluster` CRD and simply stores the TLS key and cert. The `DaskCluster` CRD would contain a reference to the secondary CRD. That way, in theory, we would have more fine-grained access control. That being said, DG would need a a service account that grants access to both CRDs anyway, and this adds substantial complexity for a small return that we may not even need. I'm a big fan of _first make it work, then make it better_. The path of least resistance, for now, is to just store it in the CRD.

droctothorpe:
> For a pod to access the secret, it must be in its own namespace. Thus, if we only store the TLS credentials in a secret, then the api server needs a cluster role to read (list, get, watch) secrets for all namespaces. This is a huge level of access, and not something that a security-minded admin might like.

@jcrist which pods need access to the TLS secrets? If it's only DG, DG will be deployed to a predetermined namespace, not a dynamically provisioned one, in which case, we can store all of the secrets in that one namespace.

jcrist:
No, the scheduler and worker pods will also need to mount their corresponding secret as a volume, and they'll be potentially running in other namespaces.

droctothorpe:
Thanks for clarifying!

> A third option would be to create the secrets in the same namespace as the DaskCluster objects, and have the controller mirror them to namespace the cluster eventually runs in. This is more complicated than storing them in the DaskCluster objects, but keeps secret things stored in secrets.

In that case, I think that this approach makes a lot of sense. There would be no need to watch the secrets in the dynamic namespaces, so no overhead, other than creation. We can make the scheduler the parent of the duplicated secret. That way, as soon as it's deleted, the duplicated secret will be deleted as well.

droctothorpe:
Scratch that, cross namespace owner references are prohibited 😢 
https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

droctothorpe:
If we make the CRD cluster-scoped, namespace-scoped operands can be owned by the CRs.

https://github.com/operator-framework/operator-sdk/blob/master/doc/operator-scope.md#crd-scope

jcrist:
I just pushed a commit updating the helm chart and reconfiguring our docker images to more cleanly build from master. I had to pull in a few other changes and rebased on master, which unfortunately probably broke others working off this branch (apologies, I'll try to avoid that in the future).

The helm chart right now:
- Deploys the gateway server
- Deploys a traefik proxy
- Adds a route to traefik for serving the gateway server
- Forwards configuration from the helm chart to the gateway server, similar to the previous version.

jcrist:
Hmmm, didn't know about the namespace limitations of ownerReferences. I don't think we should have any cluster-scoped objects, these come with additional restrictions and issues. I'd rather solve this using other methods.

Say dask-gateway (the helm chart) is installed in a namespace `dask-gateway`. The controller, api server, and proxy will all run in this namespace.

When a user creates a new cluster, we create a `DaskCluster` object in the `dask-gateway` namespace. Either we store the TLS credentials in this object, or the controller creates a `Secret` in the `dask-gateway` namespace to store the credentials (and sets the DaskCluster object as its owner) - I haven't decided which method is better yet.

When the controller is ready to start the scheduler pod, it:

- Mirrors the corresponding secret over to the target namespace (call it `cluster-namespace`).
- Creates the scheduler pod in the `cluster-namespace`, and sets the secret as the owner
- Creates a service in the `cluster-namespace` for the scheduler pod, and sets the pod as the owner
- Creates an `IngressRoute` in the `dask-gateway` namespace, and sets the `DaskCluster` object as the owner

As workers are added/removed, they're each started in the `cluster-namespace`, with the scheduler pod set as their owner.

When we want to stop a cluster we then:
- Delete the `IngressRoute` in the `dask-gateway` namespace
- Delete the `Secret` in the `dask-gateway` namespace
- Delete the `Secret` in the `cluster-namespace` and let k8s garbage collection handle the rest.

The gateway api server would need the following permissions:
- `get,list,watch,create,patch,update,delete` for `daskclusters` in the `dask-gateway` namespace
- `get,list,watch` for `secrets` in the `dask-gateway` namespace (only necessary if we don't store the credentials in the `DaskCluster` object.

The controller would need the following permissions:
- `get,list,watch,create,patch,update,delete` for `pods` in all namespaces
- `create,delete` for `services` in all namespaces
- `create,delete` for `secrets` in all namespaces
- `get,list,watch,create,patch,update,delete` for `daskclusters`, `ingressroutes`, and `secrets` in the `dask-gateway` namespace

jcrist:
One issue with the above is that pods can be orphaned if a `DaskCluster` object is ever deleted by another entity (e.g. an admin) - the gateway would lose the record and all objects in the `cluster-namespace` would be orphaned. Since the intention is that all interactions come through the api server, I expect the chance of this to be low. Orphaned objects could also be reclaimed by a periodic task to delete all `secrets` with the dask-gateway unique instance label in all namespaces that don't have a corresponding record in the `dask-gateway` namespace.

An alternative to the above would be to have `DaskCluster` objects live in the namespace they'll eventually be launched in. This makes it much easier - all child ojects (pods, etc...) are tied to their respective `DaskCluster`. Since we label all child objects with a label specific to a `dask-gateway` deployment (the helm release name), multiple gateways can run in the same cluster and not step on each other - they can filter on this label in the watch stream.

Walking through the above again:

dask-gateway (the helm chart) is installed in a namespace `dask-gateway`. The controller, api server, and proxy will all run in this namespace.

When a user creates a new cluster, the api server:

- Creates a `Secret` object in the `dask-gateway` namespace holding the TLS credentials
- Creates a `DaskCluster` object in the target namespace (call it `cluster-namespace`).

When the controller is ready to start the scheduler pod, it:

- Mirrors the corresponding secret over to the `cluster-namespace` and sets the `DaskCluster` object as its owner
- Creates the scheduler pod in the `cluster-namespace`, and sets the secret as the owner
- Creates a service in the `cluster-namespace` for the scheduler pod, and sets the scheduler pod as the owner
- Creates an `IngressRoute` in the `cluster-namespace`, and sets the scheduler pod as the owner

As workers are added/removed, they're each started in the `cluster-namespace`, with the scheduler pod set as their owner.

When we want to stop a cluster we then:
- Delete the `Secret` in the `dask-gateway` namespace
- Delete the `Secret` in the `cluster-namespace` and let k8s garbage collection handle the rest.

Deleting a `DaskCluster` object deletes all but the originating secret in the `dask-gateway` namespace. A periodic check for orphaned secrets could clear these out if they're missed.

The gateway api server would need the following permissions:
- `get,list,watch,create,patch,update` for `daskclusters` in all namespaces
- `get,list,watch,create` for `secrets` in the `dask-gateway` namespace

The controller would need the following permissions:
- `get,list,watch,create,patch,update,delete` for `pods` and `daskclusters` in all namespaces
- `create,delete` for `services`, `secrets`, and `ingressroutes` in all namespaces
- `get,list,watch` for `secrets` in the `dask-gateway` namespace

jcrist:
> I'll also sync up with our SRE team to see if they have any objections.

I talked to one of our ops people, and they don't see an issue with storing the TLS credentials in the DaskCluster object - Secrets are just a convention, and someone with access to the DaskCluster objects in a namespace can still do damage. Just one datapoint.

Right now I'm leading towards the 2nd design with credentials stored in the `DaskCluster` objects. This removes the need for `secrets` access in the api-server and keeps things simpler to implement.

droctothorpe:
I just wrapped a meeting with our SRE team. They held that creating a secret in the primary namespace where the CR lives and mirroring it to the dynamic namespace where the scheduler and workers live is the most secure and audit-proof implementation. Secrets are encrypted at rest, CRs are not. They also suggested that we base64 encode the key if we keep it in the CR. It's not as secure as using secrets, but it's better than nothing.

droctothorpe:
> Kubernetes helps keep your Secrets safe by making sure each Secret is only distributed to the nodes that run the pods that need access to the Secret. Also, on the nodes themselves, Secrets are always stored in memory and never written to physical storage, which would require wiping the disks after deleting the Secrets from them.

> ...etcd stores Secrets in encrypted form, making the system much more secure. 

> The secret volume uses an in-memory filesystem (tmpfs) for the Secret files. Because tmpfs is used, the sensitive data stored in the Secret is never written to disk, where it could be compromised.

Source: [Kubernetes in Action](https://learning.oreilly.com/library/view/kubernetes-in-action/9781617293726/kindle_split_018.html).

droctothorpe:
Here's a sample network policy:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-dask-gateway-traffic
  namespace: <dynamically provisioned namespace>
spec:
  policyTypes:
  - ingress
  - egress
  ingress:
    - from:
      - namespaceSelector:
          matchLabels:
            name: <dask gateway namespace>
         podSelector:
           matchLabels:
             name: <dask gateway pod label>
      - namespaceSelector:
          matchLabels:
            name: <dynamically provisioned namespace>
  egress:
    - from:
      - namespaceSelector:
          matchLabels:
            name: <dask gateway namespace>
         podSelector:
           matchLabels:
             name: <dask gateway pod label>
      - namespaceSelector:
          matchLabels:
            name: <dynamically provisioned namespace>
```

This ensures that the scheduler and workers can **only** communicate with each other and DG, which improves security. This doesn't account for worker data access though. We may need a more lenient ingress policy. 

Our SRE lead recommended creating the namespace and network policy **every** time a cluster is instantiated, and deleting it every time a cluster is deleted (through the `cleanup` function) . That should address the user management concern that you mentioned but may add some latency.

droctothorpe:
@jcrist per our discussion: I think that the two pod status field we're interested in are `Pod.Container.<container>.Ready` which is just a boolean, and `Pod.Container.<container>.State.Reason`, which contains a brief, pascal-case summary about what's wrong with the pod, e.g. [CrashLoopBackOff](https://github.com/kubernetes/kubernetes/blob/4fda1207e347af92e649b59d60d48c7021ba0c54/pkg/kubelet/container/sync_result.go#L29). 

Don't quote me on this, but I think that [this](https://github.com/kubernetes/kubernetes/blob/4fda1207e347af92e649b59d60d48c7021ba0c54/pkg/kubelet/container/sync_result.go#L38) is an exhaustive list of `Pod.Container.<container>.State.Reason` failure reasons. 

Also, given how the controller handles reconciliation, we may want to set the restart policy to never.

droctothorpe:
Unfortunately, there doesn't seem to be a consolidated list of potential reasons. Here's a list of some of the more common ones that I've run into before:

From lines 28 to 43.
https://github.com/kubernetes/kubernetes/blob/d24fe8a801748953a5c34fd34faa8005c6ad1770/pkg/kubelet/images/types.go#L28

https://github.com/kubernetes/kubernetes/blob/4fda1207e347af92e649b59d60d48c7021ba0c54/pkg/kubelet/container/sync_result.go#L29

https://github.com/kubernetes/kubernetes/blob/4fda1207e347af92e649b59d60d48c7021ba0c54/pkg/kubelet/container/sync_result.go#L38

https://github.com/kubernetes/kubernetes/blob/4fda1207e347af92e649b59d60d48c7021ba0c54/pkg/kubelet/container/sync_result.go#L40

jcrist:
I've just pushed a pass at implementing the controller in python here. This removes all the operator-sdk things and does everything in Python. I like this approach better - dask developers are generally more familiar with Python than go, and we get to reuse much of the infrastructure we already have in Python.

In its current state the controller & backend successfully manages to start a scheduler, add it to the proxies, connect the user, and shutdown the cluster, all while managing failures. There are likely still bugs, but it's a start. Next step is to add worker support.

yuvipanda:
YESSSSSSS for implementing the controller in Python :)

jcrist:
Status update:

- Handling of the scheduler (pod, service, ingresses, etc...) is now more robust to failures and delays. Specifically, it's tricky to know when the route is *available* on the proxies, since there's no feedback from traefik through the k8s api (no status field on their CRDs). To get around this we delay creation of the ingressroute objects until everything is likely to succeed quickly (pod running and ready, endpoint already created) and hope for the best. The client already retries on connect failures, but it retries in a really tight loop (possibly due to a bug in tornado) - it would be good to fix this to backoff at a reasonable rate.

- Scale and adapt requests are now properly forwarded from client -> proxy -> scheduler -> api server -> updating replicas in DaskCluster k8s object. The controller doesn't handle these updates yet, that's next.

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -342,35 +380,45 @@ def get_cluster_key(self, obj):
...
-                        cluster, sched_pod
-                    )
-                    status_update["ingressroutetcp"] = route
-
-                status_update["phase"] = "Running"
+                endpoints = self.informers["endpoints"].get(
+                    f"{namespace}.{service_name}"
+                )
+                if endpoints is not None and self.endpoints_all_ready(endpoints):
+                    # Only add routes if endpoints exist, otherwise traefik gets grumpy
```
droctothorpe: Just out of curiosity, what do you mean by `grumpy`?

droctothorpe:
Hey Jim, can you confirm that waiting for a pod to be ready before creating an ingress is non-blocking?

For example, what if a user calls `new_cluster` but the Kubernetes worker nodes have no more room for the scheduler pod, so a new Kubernetes worker node needs to spin up and join the cluster, a process which can take several minutes, then a second user scales their cluster down (through a client update to their  custom resource) -- will the second user's action be blocked? 

Also, what will happen in the client in the event that the pod isn't ready for several minutes?

jcrist:
Ok, this is now working. Running `skaffold run` should deploy everything in a test cluster (I've been using k8s on docker-desktop). There are still a few things it needs:

- Tests. I have some simple "Is this configuring things properly" tests, but need to add tests for the actual controller and backend logic. I'd also like to test our deployment infrastructure (helm), which will probably be just testing the templates as well as an integrated deployment.
- Some error handling could be better, particularly in the presence of resource quotas. We should be robust to these errors, but the errors will be seen as any-other-error (rather than something expected like a resource-quota). We should backoff and try to resubmit the pods later.
- Pods currently are deployed in one batch. If a user scales a cluster up to 100 workers, we submit 100 pods immediately. We should probably ramp up in batches to avoid overloading api server, as well as to avoid doing unnecessary work in the presence of resource quotas (submitting 100 pods and having them all be denied by a resource quota is more expensive than failing earlier in small batches).

There are also definitely bugs. If anyone is feeling adventurous and wants to try deploying and using this PR, poking at it and reporting issues would be useful. The helm chart values file as well as the values I've been using for local deployment (in `resources/helm/testing/skaffold.yaml`) are all the documentation we have right now, but I'd be happy to help get someone set up.

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
jcrist: It just logs a lot of errors, everything still works fine. We also want to delay moving the cluster into the running state until the route is likely to succeed being added to the proxy (unfortunately we can't be *sure* it's added unless we ping the proxy (which I don't want to do), but if the endpoint already exists, adding it to the proxy should happen almost immediately).

jcrist:
> Hey Jim, can you confirm that waiting for a pod to be ready before creating an ingress is non-blocking?
>
> For example, what if a user calls new_cluster but the Kubernetes worker nodes have no more room for the scheduler pod, so a new Kubernetes worker node needs to spin up and join the cluster, a process which can take several minutes, then a second user scales their cluster down (through a client update to their custom resource) -- will the second user's action be blocked?

No, everything happens asynchronously. Both operations can happen simultaneously. It also won't block the controller since everything is event based (there is no task in the controller waiting for the pod to start).

> Also, what will happen in the client in the event that the pod isn't ready for several minutes?

Calling `gateway.new_cluster()` will block until the cluster is ready. If that takes a few minutes that's fine - the client opens a series of long polling requests while waiting (which either end after a timeout, or the cluster is known to be running or stopped/failed). Waiting for a cluster to start isn't expensive for either the server or client.

droctothorpe:
I'm OOO tomorrow but really look forward to deploying and testing on Wednesday.

jcrist:
Todo list:

- [ ] Tests
- [ ] Cleanup logging of operations, we're not logging everything that's important, and the logging levels and verbosity isn't consistent everywhere.
- [ ] Add `stop_time` timestamp to model (we're already setting `start_time`).
- [ ] Delete stopped/failed clusters after a timeout. Currently these accumulate and must be deleted manually or with a cronjob.
- [ ] Expose backoff and rate limiting knobs in the configuration options. These are currently hardcoded.
- [ ] Expand docstrings and comments

gforsyth:
Hey @jcrist -- this is looking great!  We just deployed this to try it out.  We needed to tweak a few small things in the `helm` chart for enterprise-y stuff that we can push up later, but I don't think that's a blocker for public consumption.

We can hit the gateway, log in, `list_clusters` just fine.  `create_cluster` shows a pod being created (and it is running) but the call to `create_cluster` on the client side never returns and the `controller` is showing this error:

```
Traceback (most recent call last):
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/controller.py", line 437, in reconciler_loop
    requeue = await self.reconcile_cluster(name)
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/controller.py", line 457, in reconcile_cluster
    status_update, requeue = await self.handle_cluster(cluster)
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/controller.py", line 487, in handle_cluster
    return await self.handle_pending_cluster(cluster)
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/controller.py", line 506, in handle_pending_cluster
    cluster
  File "/opt/conda/lib/python3.7/site-packages/dask_gateway_server/backends/kubernetes/controller.py", line 790, in create_scheduler_pod_if_not_exists
    return pod["metadata"]["name"], pod
TypeError: 'NoneType' object is not subscriptable
```

jcrist:
> shows a pod being created (and it is running) but the call to create_cluster on the client side never returns and the controller is showing this error:

Ah, I can see how the bug here, but I'm curious why it's occurring on your side and not on mine. Can you post the full controller logs?

jcrist:
Should be fixed now.

gforsyth:
> Ah, I can see the logic bug here, but I'm curious why it's occurring on your side and not on mine. Can you post the full controller logs?

Sorry, changes were pushed up to come CRDs and I think our issue was that some CRDs were missing permissions (a lot of this, I believe, is due to security restrictions on our end about how verbs are listed, etc.)

jcrist:
> Sorry, changes were pushed up to come CRDs and I think our issue was that some CRDs were missing permissions (a lot of this, I believe, is due to security restrictions on our end about how verbs are listed, etc.)

Did you need to change the service account permissions for the controller/api server? The permissions created by the helm chart should be state the permissions required explicitly - if these aren't sufficient then this is a bug in our helm chart. I'd be interested in seeing what you needed to do to get things working.

droctothorpe:
One notable change was this in the controller RBAC:
```
rules:
--
  | - apiGroups: ["gateway.dask.org"]
  | resources: ["daskclusters", "daskclusters/status"]
  | verbs: ["get", "create", "list", "watch", "delete", "patch", "update"]
```
`daskclusters` and `daskclusters/status` have discrete permissions.

jcrist:
Hmmm, didn't know that `*/status` things were separate resources. That should have caused things to fail on any system with rbac enabled, but it looks like docker-desktop automatically promotes all service accounts to admin by default (see https://github.com/docker/for-mac/issues/3694#issuecomment-498254793) - deleting their clusterrolebinding causes things to fail properly locally as well.

jcrist:
I've fixed the RBAC issue. I also updated the helm chart to serve over a single port by default, as this will likely be the more common configuration.

jcrist:
This is getting pretty close to ready to merge. I'm going to try and expand unit tests tomorrow, and then hit the merge button. I think getting larger integration tests setup will be a subsequent PR - I've at least verified things work locally and this PR is pretty big already.

droctothorpe:
You may find OSDKs documentation on E2E tests to be a helpful reference: https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md

droctothorpe:
Do other backends rely on the Golang proxy that you wrote?

jcrist:
Yes. All other backends use our bundled proxy, the kubernetes backend is the only one using traefik.

jcrist:
Ok, I'm going to merge this.

Since the design has changed a few times during this PR (and some of the above comments are out-of-date), here is an up-to-date summary of what was implemented here:

- A `DaskCluster` CRD, storing a spec and status for each dask cluster
- A `Backend` implementation for kubernetes, which stores all state in `DaskCluster` objects 
- A controller, implemented in Python, which watches for dask cluster objects and processes them accordingly.
- A helm chart for deploying the gateway api server, controller, and traefik proxy together, configured appropriately
- Updated docker images that build from source rather than pypi for easy deployment from source
- A skaffold setup that can optionally run the controller out-of-cluster (by specifying `-p local-controller`) for easier development

There are currently some unit tests, mostly that k8s objects created have the appropriate fields and that configuration is forwarded appropriately. I've poked at things locally and everything seems ok. Integration tests will be added in a subsequent PR.

PR #207, opened 2020-02-18
Add Skaffold configuration and readme

gforsyth:
This PR adds in a configuration for using a local k8s development workflow using Skaffold.  Instructions for how to get started and to use `skaffold` are in `SKAFFOLD.MD`.

This has been tested with docker-desktop kubernetes but should work with whatever the currently active k8s context is.


jcrist:
Thanks! Is there a way this could be added to not the root directory? I might put this under either `resources/helm/skaffold`, or something new like `build_tools/skaffold/`.

This also might be self-contained enough to merge against master. If you agree, you can change the target branch using the github ui, see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request.

gforsyth:
Hey @jcrist -- I was just thinking that.  We can move everything _except_ the `skaffold.yaml` file into a `skaffold` subdirectory (otherwise we don't have context when we copy the `dask-gateway-server` src directory during the build)

jcrist:
Sure, makes sense to me. Just trying to de-clutter the root directory.

gforsyth:
Ready for another look @jcrist !  And I agree that this is self-contained enough so I've retargeted it to `master`

droctothorpe:
🤔 When the operator is a little further along, we may want to update `skaffold.yaml` such that it tracks and redeploys the operator source code as well. Until then, the operator can be run on the host using `operator sdk run --local --operator-flags="--zap-encoder=console"'` and Skaffold can be used to automate the development cycle of Dask Gateway exclusively.

jcrist:
This looks good enough to me. I'm going to merge this now and try it out while I update the helm chart in the kube-support branch. Thanks!

PR #208, opened 2020-02-18
Trivial change to test CI

jcrist:
There's a weird unrelated (I think) failure showing up in #206, debugging it here.


jcrist:
Yep, tests still hang on master. I'm unable to reproduce this locally. I suspect this has to do with an upstream dependency change (possible dask?). Will continue debugging here.

jcrist:
Fixed by https://github.com/dask/distributed/pull/3494, closing.

PR #209, opened 2020-02-20
Update CR to use PodTemplateSpec and add compatible CR

droctothorpe:
@jcrist Here's the PR that updates the custom resouce definition to use a PodTemplateSpec instead of a PodTemplate. I've also included a little bit of documentation to facilitate local development.


jcrist:
LGTM, thanks!

PR #210, opened 2020-02-20
Remove versioneer

jcrist:
Previously we used versioneer to manage versioning based on git tags.  I
like versioneer, I use it for many projects, but it doesn't seem to be a
good fit here:

- This repo contains multiple python packages. Using versioneer means
that they have to be released in tandem, there's no way to increment a
version of one without incrementing the version of the other.

- Building the package requires either a git-produced tarball, or a
`.git/` directory to extract the version info from. This makes building
the package from source tricky in other contexts (e.g. docker) since we
don't always want to ship the full `.git` folder into the docker context
(it may be large!).

To get around this, we switch to defining the version manually. Using a
bit of trickery we can avoid having to define the version of each
package in two places (`setup.py` and `__init__.py`), but it will still
need to be incremented in code during a release. I'm fine with this,
this project is tricky enough to release as is.


gforsyth:
Not sure if you've played with `rever` before, but it might be helpful.  https://regro.github.io/rever-docs/

jcrist:
I haven't, looks interesting though, thanks. If we decide we need more automation later we may look into it.

PR #211, opened 2020-02-21
[WIP] Preliminary controller implementation

droctothorpe:


droctothorpe:
@jcrist Thanks for being patient.

droctothorpe:
@jcrist Just pushed up an `active` field. When set to false, the controller deletes the scheduler pod, which in turn deletes the worker pods, since they're owned by the scheduler pod. DG will want to set this field to `true` on CR creation. Happy to rename / adjust it.

jcrist:
Apologies, I rebased the kube-support branch which resulted in merge conflicts :/ (will avoid doing that in the future). Thanks for doing this!

jcrist:
Hey, so I gave this a read through and I'm thinking that the osdk actually doesn't give us that much of a benefit over writing this in Python ourselves. There's also the language barrier - dask developers are generally pretty good at Python, but may not have any Go experience. Mind if I write up a POC and we can talk about it at the dask-summit later this week?

droctothorpe:
Happy to follow your lead, @jcrist. Should I hold off on further contributions to the Golang implementation until the Python POC is complete?

jcrist:
If you wouldn't mind, yeah. I'd like to compare the implementations before having you push further on this work. Thanks for being understanding!

jcrist:
Since we're going with the controller written in Python in #206, I'm going to close this. Thanks for your work on this, seeing how it would be done in Go was helpful.

PR #225, opened 2020-03-19
Allow use of existing cluster-roles

### LINKED ISSUE #223: Allow for separate `create` vs `enable` step for `ClusterRole` and `ServiceAccount`
I'm going to try to get to this tomorrow but leaving a note here for myself and others.

In an enterprise setting, SREs can lock down the creation of cluster-wide resources which prevents the creation of new cluster roles.  In this instance, there's a need to distinguish between `create` and `enable` for things like `rbac`.  There can be a `create: False, enable: True` scenario where we want to create the SA and have it use the appropriate `ClusterRole` (created by other means / admins). 

This can be accomplished by adding a few extra branches in the `rbac` templates.

gforsyth:
Fixes #223

### DIFF
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -201,3 +201,14 @@ traefik:
 rbac:
   # Whether to enable RBAC.
   enabled: true
+
+  # Existing names to use if cluster roles have already been created by other
+  # means (leave set to `null` to create all required roles at install time)
+  controller:
+    serviceAccountName: null
+
+  gateway:
+    serviceAccountName: null
+
+  traefik:
+    serviceAccountName: null
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -202,8 +202,9 @@ rbac:
   # Whether to enable RBAC.
   enabled: true
 
-  # Existing names to use if cluster roles have already been created by other
-  # means (leave set to `null` to create all required roles at install time)
+  # Existing names to use if ClusterRoles, ClusterRoleBindings, and
+  # ServiceAccounts have already been created by other means (leave set to
+  # `null` to create all required roles at install time)
   controller:
     serviceAccountName: null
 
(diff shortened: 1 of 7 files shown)


resources/helm/dask-gateway/values.yaml
```diff
@@ -201,3 +201,14 @@ traefik:
...
+  # Existing names to use if cluster roles have already been created by other
+  # means (leave set to `null` to create all required roles at install time)
+  controller:
+    serviceAccountName: null
+
+  gateway:
+    serviceAccountName: null
+
+  traefik:
+    serviceAccountName: null
```
jcrist: Is there a good reason to include these here rather than in the existing `controller`/`gateway`/traefik` groups? My initial thought was something more like:

```yaml
rbac:
  enable: true

controller:
  serviceAccountName: null
  ...

gateway:
  serviceAccountName: null
  ...

traefik:
  serviceAccountName: null
  ...
```

gforsyth: yeah, that's a better idea

gforsyth:
Github weirdly isn't letting me comment inline in response to your first comment:

I was a bit sloppy with language earlier -- also, I'm biased by setups I've worked with.  The creation of a service account to _use_ a cluster role isn't often restricted (in my experience) -- just the cluster-wide resources, hence the break I put in.  Happy to do it either way, with a slight preference for creation of SAs at install time.

gforsyth:
Although, thinking about it, the name of the SA has to match the cluster role anyway, so doing it my way makes it less flexible.

jcrist:
IMO a `serviceAccountName` is a nice place to break out resources to be managed by the admins. The chart either creates the service account (and all cluster roles, role bindings, etc...), or it uses the referenced service account. Having the chart still create the service account feels more confusing to me (admins are responsible for binding cluster roles to the service account with a specific name, but not creating that service account, which feels odd).

gforsyth:
Looking over the changes, since the creation or not of the service accounts is all related to `rbac` being enabled or not, I think it might be a better grouping to put them all under the top-level `rbac` directive -- proximity in configuration file as a bit of a guide

jcrist:
Sure, I could go either way on that - both make sense to me.

gforsyth:
Hey @jcrist - sorry for the delay, this is ready for another look.

jcrist:
Looks good to me, thanks. Merging.

PR #227, opened 2020-03-19
Run traefik as non-root user

### LINKED ISSUE #224: Add `PodSecurityPolicy` directive for traefik pod
The traefik default pod runs as root for Reasons™: https://github.com/containous/traefik-library-image/issues/38

If a `PodSecurityPolicy` is in place, it will prevent `traefik` from starting (because `root`) -- fix is to add an option to `runAsUser` to the `traefik` spec if `PodSecurityPolicy` is enabled.

jcrist:
The traefik image defaults to running as root. This can cause issues if
a `PodSecurityPolicy` is enabled. To avoid this, we explicitly set a
`securityContext` on the traefik container, running as user `1000:1000`.

Fixes #224.


jcrist:
cc @gforsyth.

PR #230, opened 2020-03-25
Fix for dashboard proxying with multi-namespace deployment

gforsyth:
Setting `gateway.backend.namespace` to a non-null value makes clusters deploy in
that namespace, but breaks the traefik proxying of the dashboards.  Deploying
the middleware to the backend namespace makes everything happy again.

### DIFF
--- a/resources/helm/dask-gateway/templates/controller/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/controller/configmap.yaml
@@ -13,7 +13,8 @@ data:
     c.KubeController.gateway_instance = '{{ include "dask-gateway.fullname" . }}'
     c.KubeController.proxy_prefix = "{{ .Values.gateway.prefix }}"
     c.KubeController.proxy_web_middlewares = [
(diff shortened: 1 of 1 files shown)


jcrist:
Thanks Gil. I was hoping (but forgot to check/test clearly) that middlewares weren't namespaced in traefik, but that appears not to be the case. Since we can have clusters in multiple namespaces (as well as the gateway api server, which also uses this middleware) we'll need to come up with a deeper solution. I hope there's a way we can successfully reference a middleware created in one namespace from multiple namespaces - the full traefik CRD docs are here, there might be something there we can use: https://docs.traefik.io/routing/providers/kubernetes-crd/. Otherwise we'll need to figure something else out.

I am without a computer for at most the next week-and-a-half, but will look into this when I can (if you don't get to it first).

gforsyth:
I think I have a general fix for this.  Testing locally then I'll push it up.

gforsyth:
@jcrist -- no rush on this, but I pushed up a fix for this that generalizes to any number of namespaces (just a small tweak to the `configmap` was needed to specify where the `middleware` lived)

jcrist:
Ah cool, thanks! Makes sense to me, merging.

PR #231, opened 2020-03-31
Fix link to podspec in traitlet description

gforsyth:
Kubernetes spec links for `scheduler_extra_pod_config` and
`worker_extra_pod_config` were pointing at the container spec links
instead of podspec link


jcrist:
Thanks!

PR #234, opened 2020-04-07
Add user-level configurable annotations and labels

gforsyth:
This adds two new configurable options for the `KubeClusterConfig`:
`extra_annotatations` and `extra_labels`. These can be configured using
`cluster_options` and get merged with `common_annotations` and
`common_labels`, respectively.

There are already config options for `common_labels` and
`common_annotations`, so why does this exist?  Those config options are
part of the `KubeBackend` and can be used by a cluster admin to add an
extra label (say) to every pod Dask Gateway spins up.

`extra_annotations` and `extra_labels` can be configured on a
user-by-user basis, to allow for attaching user-specific annotations to
help configure data access (as one example). These options can be
populated using external services based on information in the `User`
model (name, groups, admin).

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -234,7 +234,6 @@ class KubeClusterConfig(ClusterConfig):
 
         This dict will be merged with ``common_annotations`` before being
         applied to user pods.
-
         """,
         config=True,
     )
@@ -250,7 +249,6 @@ class KubeClusterConfig(ClusterConfig):
 
         This dict will be merged with ``common_annotations`` before being
         applied to user pods.
-
         """,
         config=True,
     )
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -254,6 +254,7 @@ class KubeClusterConfig(ClusterConfig):
         config=True,
     )
 
+
 class KubeBackendAndControllerMixin(LoggingConfigurable):
     """Shared config between the backend and controller"""
 
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -223,10 +223,9 @@ class KubeClusterConfig(ClusterConfig):
         config=True,
     )
 
-    extra_annotations = Dict(
+    worker_extra_pod_annotations = Dict(
         help="""
-        Any extra annotations to be applied to a user's scheduler and worker
-        pods.
+        Any extra annotations to be applied to a user's worker pods.
 
         These annotations can be set using with cluster options (See
         :ref:`exposing-cluster-options`) to allow for injecting user-specific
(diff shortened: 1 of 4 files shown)


jcrist:
Hmmm, I'm wondering if we'd ever want object-specific labels/annotations, rather than a general set applied to all objects. For reference, the types of objects we create for a cluster are:

- pods (both scheduler and worker)
- service (for the scheduler pod only)
- secret (tls credentials)
- ingressroute & ingressroutetls (traefik routes)

Are there labels/annotations we'll want to apply to one object but not others? It's certainly simpler to apply to all objects, but may not be flexible enough.

gforsyth:
> Are there labels/annotations we'll want to apply to one object but not others? It's certainly simpler to apply to all objects, but may not be flexible enough.

I'm not certain, but in the interest of being forward looking we could do:
`extra_pod_annotations` and `extra_pod_labels` (leaving room to allow for `service`, `secret`, etc.)

I'm trying to think, also, if we would ever want specific extra labels/annotations for workers vs. the scheduler

jcrist:
Hmmm, I'd rather keep things as simple as possible for now, and add extra options as needed. Can you say a bit more about what your needs are? What objects specifically do you want to add extra labels/annotations to?

gforsyth:
We only need extra annotations and labels on pods

jcrist:
Oh, that makes things a little easier. For the sake of consistency with the `scheduler_extra_*` and `worker_extra_*` config fields, lets go with:

```
scheduler_extra_pod_labels
scheduler_extra_pod_annotations
worker_extra_pod_labels
worker_extra_pod_annotations
```

gforsyth:
@jcrist updated using the separate worker / scheduler names for both labels and annotations.

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -1149,10 +1153,17 @@ def make_pod(self, namespace, cluster_name, config, is_worker=False):
         if extra_container_config:
             container = merge_json_objects(container, extra_container_config)
 
+        annotations = self.common_annotations
+        if extra_pod_annotations:
+            annotations = merge_json_objects(annotations, extra_pod_annotations)
```
jcrist, line 1158: Annotations can only be a map, this could be:

```python
annotations = self.common_annotations.copy()
annotations.update(extra_pod_annotations)
```

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -1149,10 +1153,17 @@ def make_pod(self, namespace, cluster_name, config, is_worker=False):
...
+
+        if extra_pod_labels:
+            labels.update(extra_pod_labels)
```
jcrist, line 1161: No need for conditional here, `extra_pod_labels` will always be a dict (it just might be empty).

jcrist: Since `labels` isn't used before then, you might also move https://github.com/dask/dask-gateway/blob/5377c03875682cea760c5ce61add8b646ec170aa/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py#L1120 down to here to better group the logic.

dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
```diff
@@ -223,8 +223,68 @@ def _default_scheduler_memory_limit(self):
...
+        Any extra annotations to be applied to a user's scheduler pods.
+
+        These annotations can be set using with cluster options (See
+        :ref:`exposing-cluster-options`) to allow for injecting user-specific
+        information, e.g. adding an annotation based on a user's group or
+        username.
+
+        This dict will be merged with ``common_annotations`` before being
+        applied to user pods.
+
```
jcrist, line 253: Unneeded blank line.

dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
```diff
@@ -223,8 +223,68 @@ def _default_scheduler_memory_limit(self):
...
+        Any extra annotations to be applied to a user's worker pods.
+
+        These annotations can be set using with cluster options (See
+        :ref:`exposing-cluster-options`) to allow for injecting user-specific
+        information, e.g. adding an annotation based on a user's group or
+        username.
+
+        This dict will be merged with ``common_annotations`` before being
+        applied to user pods.
+
```
jcrist, line 237: Unneeded blank line.

gforsyth:
> Note that if you want the k8s tests to run on CI you need to include the string `[test-kube]` somewhere in your commit.

Whoops.  Thanks, I had _definitely_ missed that.  Addressed your other comments -- thanks for the speedy review!

jcrist:
> Whoops. Thanks, I had definitely missed that.

No worries. I'm not sure how to make this process cleaner. We don't want to test all backends on every PR because that may be expensive. Perhaps adding this info to a PR template? 🤷.

Anyway, LGTM, merging.

PR #241, opened 2020-04-09
Fix k8s helm chart support for prefixed deployments 

### LINKED ISSUE #239: Prefix config in 0.7.0, dask-gateway-server won't serve under it
It is possible to configure a prefix in the Helm chart. But I think the dask-gateway-server binary running in `deploy/api-<release>` isn't getting configured to expect receiving such traffic from `deploy/traefik-<release>`.

## Example

The JupyterHub Helm chart allow us to register a JupyterHub service through its configuration. During this registration, we can configure `url` that will make JupyterHub configure its proxy to route `/services/<service-name>` to that url for us. This way we can reuse JupyterHub's proxy's ability to manage HTTPS traffic and remain on the same domain etc.

#### jupyterhub/config.yaml (Helm values)
```yaml
hub:
  services:
    # ref: https://jupyterhub.readthedocs.io/en/stable/reference/services.html#properties-of-a-service
    dask-gateway:
      apiToken: "redacted"
      url: http://traefik-dask-gateway.dask-gateway
```

#### dask-gateway/config.yaml (Helm values)
```yaml
gateway:
  loglevel: INFO
  # Path prefix to serve dask-gateway api requests under
  # This prefix will be added to all routes the gateway manages
  # in the traefik proxy.
```

## Issue summary

From logs to deploy/api I can see that it receives traffic if I curl `myhub.com/services/dask-gateway/api/version` to the path `/services/dask-gateway/api/version`, but it return 404 to it but responds properly to `/api/version`. So, the `gateway.prefix` config seem only to configure treafik to direct traffic properly, but fail to configure the dask-gateway-server binary in `deploy/api` to expect traffic on that path.

```
(base) ➜  jupyterhub git:(master) ✗ kubectl logs -n dask-gateway deploy/api-dask-gateway
[I 2020-04-09 10:49:03.515 DaskGateway] Starting dask-gateway-server - version 0.7.0
[I 2020-04-09 10:49:03.793 DaskGateway] Authenticator: 'dask_gateway_server.auth.JupyterHubAuthenticator'
[I 2020-04-09 10:49:03.794 DaskGateway] Backend: 'dask_gateway_server.backends.kubernetes.backend.KubeBackend'
[I 2020-04-09 10:49:03.819 DaskGateway] Dask-Gateway server started
[I 2020-04-09 10:49:03.820 DaskGateway] - Private API server listening at http://:8000
[I 2020-04-09 10:49:14.516 DaskGateway] 404 GET /services/dask-gateway/api/version 0.661ms
[I 2020-04-09 10:54:51.444 DaskGateway] 200 GET /api/version 0.646ms
```

## Questions

- Is there a configuration that we can write to the configuration of dask-server-gateway that would help it serve content under the prefix?
- We could also make traefik rewrite the path to `/` before sending traffic onwards to the dask-gateway-server, but sometimes applications needs to know where the client is accessing it from. If this is the case dask-gateway-server, this is a bad option.

jcrist:
Fixes support for deploying dask-gateway on k8s with all routes served
under a prefix (e.g. `/services/dask-gateway/`).  This was woefully
broken on master, and due to lack of tests was missed.

Fixes #239


jcrist:
cc @consideratio, this should fix your issue.

resources/helm/dask-gateway/templates/gateway/middleware.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ include "dask-gateway.fullname" . | printf "clusters-prefix-%s" | trunc 63 | trimSuffix "-" }}
+  labels:
+    {{- include "dask-gateway.labels" . | nindent 4 }}
+spec:
+  stripPrefixRegex:
+    regex:
+      - '{{ .Values.gateway.prefix | trimSuffix "/" }}/clusters/[a-zA-Z0-9.-]+/'
```
consideRatio, line 10: This middleware will not work if someone calls `/clusters/name1`, but only if someone calls `/clusters/name1/`. This is a small difference that I've recently run into issues with for other software.

I'm considering this:

```suggestion
      - '{{ .Values.gateway.prefix | trimSuffix "/" }}/clusters/[a-zA-Z0-9.-]+'
```

And if we should make things more explicit by setting this: https://docs.traefik.io/middlewares/stripprefix/#forceslash

resources/helm/dask-gateway/templates/gateway/middleware.yaml
```diff
@@ -0,0 +1,23 @@
...
+{{- if ne .Values.gateway.prefix "/" }}
+---
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: {{ include "dask-gateway.fullname" . | printf "api-prefix-%s" | trunc 63 | trimSuffix "-" }}
+  labels:
+    {{- include "dask-gateway.labels" . | nindent 4 }}
+spec:
+  stripPrefixRegex:
```
consideRatio, line 20:
```suggestion
  stripPrefix:
```

resources/helm/dask-gateway/templates/gateway/middleware.yaml
```diff
@@ -0,0 +1,23 @@
...
+    regex:
+      - '{{ .Values.gateway.prefix | trimSuffix "/" }}/'
```
consideRatio, line 22: Similarly as above, I consider:

```suggestion
      - '{{ .Values.gateway.prefix | trimSuffix "/" }}'
```

jcrist:
Thanks for the comments, updated. Note that none of the backends (api server or cluster dashboards) support a root route, so forcing a trailing slash isn't a problem here. Removed it anyway since it causes no harm.

consideRatio:
Ah! A concern I had was that the rule would not trigger on a request that didn't include the trailing `/`.

Thanks for your quick work to resolve this!

jcrist:
If you're able to deploy from master, I'm going to try to fix your other issue and push a patch release in the next few days. Please let me know if you run into any other problems.

PR #256, opened 2020-04-17
Add liveness and readiness probes to API deployment

droctothorpe:
Idiomatically, production pods should leverage liveness probes. Liveness probes ensure that pods are restarted if the application is unresponsive but the main process is still running. 

> For pods running in production, you should always define a liveness probe. Without one, Kubernetes has no way of knowing whether your app is still alive or not. As long as the process is still running, Kubernetes will consider the container to be healthy.
-- From **Kubernetes in Action** by Marko Luksa

Dask Gateway Server includes a [health check endpoint](https://github.cloud.capitalone.com/dask/dask-gateway/blob/bfb771cec733571bf9090349316cfce6ca9d1b1d/dask-gateway-server/dask_gateway_server/routes.py#L70), but kubelet isn't checking it. 

This PR updates the Helm deployment for Dask Gateway Server to include [liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/). 

I tested this on my local machine using Docker Desktop Kubernetes and confirmed that everything works. 

Major kudos to @suchitpandya for identifying this discrepancy.

### DIFF
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -57,19 +57,25 @@ spec:
           ports:
             - containerPort: 8000
               name: api
+
+          {{- if .Values.gateway.livenessProbe.enabled }}
           livenessProbe:
             httpGet:
               path: /api/health
               port: api
-            initialDelaySeconds: 5
-            periodSeconds: 10
-            timeoutSeconds: 2
-            failureThreshold: 3
+            initialDelaySeconds: {{ .Values.gateway.livenessProbe.initialDelaySecond }}
+            periodSeconds: {{ .Values.gateway.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.gateway.livenessProbe.timeoutSeconds }}
+            failureThreshold: {{ .Values.gateway.livenessProbe.failureThreshold }}
+          {{- end }}
+
+          {{- if .Values.gateway.livenessProbe.enabled }}
           readinessProbe:
             httpGet:
               path: /api/health
               port: api
-            initialDelaySeconds: 5
-            periodSeconds: 10
-            timeoutSeconds: 2
-            failureThreshold: 1
+            initialDelaySeconds: {{ .Values.gateway.readinessProbe.initialDelaySecond }}
+            periodSeconds: {{ .Values.gateway.readinessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.gateway.readinessProbe.timeoutSeconds }}
+            failureThreshold: {{ .Values.gateway.readinessProbe.failureThreshold }}
+          {{- end }}
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -57,3 +57,19 @@ spec:
           ports:
             - containerPort: 8000
               name: api
+          livenessProbe:
+            httpGet:
+              path: /api/health
+              port: api
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 2
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /api/health
+              port: api
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 2
+            failureThreshold: 1
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -63,7 +63,7 @@ spec:
             httpGet:
               path: /api/health
               port: api
-            initialDelaySeconds: {{ .Values.gateway.livenessProbe.initialDelaySecond }}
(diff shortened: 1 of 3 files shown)


resources/helm/dask-gateway/templates/gateway/deployment.yaml
```diff
@@ -57,3 +57,19 @@ spec:
...
+            timeoutSeconds: 2
+            failureThreshold: 3
+          readinessProbe:
+            httpGet:
+              path: /api/health
+              port: api
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            timeoutSeconds: 2
+            failureThreshold: 1
```
droctothorpe, line 75: The `readinessProbe` fails first, which cuts traffic to the impacted pod but doesn't restart it, giving the application time to recover. After 3 failed health checks (line 67), the recovery window closes, the `livenessProbe` is triggered, and the pod is replaced.

droctothorpe:
Happy to contribute in whatever small way I can. Updated per your specifications and validated:
image

resources/helm/dask-gateway/values.yaml
```diff
@@ -54,6 +54,23 @@ gateway:
...
+    periodSeconds: 10
+    failureThreshold: 3
+  readinessProbe:
+    # Enables the readinessProbe.
+    enabled: true
+    # Configures the readinessProbe.
+    initialDelaySeconds: 5
+    timeoutSeconds: 2
+    periodSeconds: 10
+    failureThreshold: 1
```
jcrist, line 72: Are these good defaults? I think the period and timeout seem fine, but am a bit concerned about how small the `failureThreshold` is for readiness. The default is 3, is there a good reason to use 1 here? The `initialDelaySeconds` also might be a bit small for liveness, if the process is slow to start it only has 35 seconds before it will be killed and restarted - should we worry about that?

I do agree that the default time ratio between failing readiness and liveness should be 3, that seems reasonable. Thoughts @droctothorpe?

droctothorpe: I'm happy to substitute with other values. My research into sensible defaults wasn't exhaustive. I also consulted with a member of our SRE team before setting them.

Here's some of the reasoning:
* `initialDelaySeconds` -- AFAIK, this doesn't start until the container is already up and running. Since everything is preconfigured, the actual process start time is really fast. I even tested a 3 second `initialDelaySeconds` and ran into no issues. 
* livenessProbe `failureThreshold` is pretty low because if the API can't respond to a health check, it probably can't respond to other kinds of traffic. `2` might be a reasonable alternative if you want it to keep sending traffic for another 10 seconds. Also, if you have a second API deployed (which is advisable in a production environment), you'd want to start routing traffic to the healthy pod rather than the unhealthy one sooner rather than later. 
* `failureThreshold` is set to 3 (with a `periodSeconds` of 10 for a total of 30 seconds of failed health checks) because, by that point, the API has been unresponsive for 30 seconds and it only takes a few seconds to reap and replace the pod. Fundamentally, the pods are disposable and can be rapidly replaced. 

Totally open to alternatives.

jcrist: I think this makes sense for a production deployment (with `replicas > 1`). Since the default is 1 replica, and many users won't change that, I think we want to be less aggressive by default.

Perhaps `failureThreshold` of `3` for `readinessProbe`, and `6` for `livenessProbe`? That keeps the time delay between failing readiness and liveness, but also adds some slack for users running the chart unmodified.

droctothorpe: Copy that. Pushed the update.

jcrist:
LGTM. Thanks!

dhirschfeld:
I'm not an expert (by any means) but I've read that you have to be cautious with using liveness probes:
https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html

Not sure if it's relevant here (prob depends on the implementation of the `/api/health` endpoint), but perhaps worth keeping in mind...

jcrist:
Yeah, the important thing is that checks for the liveness probe don't rely on any *downstream services*, they only check the health of the pod itself. Otherwise you run the risk of cascading failures. We're fine here in this regard.

resources/helm/dask-gateway/templates/gateway/deployment.yaml
```diff
@@ -57,3 +57,25 @@ spec:
...
+
+          {{- if .Values.gateway.livenessProbe.enabled }}
+          livenessProbe:
+            httpGet:
+              path: /api/health
+              port: api
+            initialDelaySeconds: {{ .Values.gateway.livenessProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.gateway.livenessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.gateway.livenessProbe.timeoutSeconds }}
+            failureThreshold: {{ .Values.gateway.livenessProbe.failureThreshold }}
```
consideRatio, line 69: To reduce the issues of working with Helm's templates, I suggest opting for doing something more like.

```yaml
{{- with .Values.gateway.livenessProbe }}
livenessProbe:
  {{- .Values.gateway.livenessProbe | toYaml | nindent 12 }}
{{- end }}
```

PR #257, opened 2020-04-22
Fix dashboard_url for distributed.Client

### LINKED ISSUE #247: Client repr of dashboard_url doesn't take GatewayCluster into account
The repr of a `distributed.Client` includes the dashboard url, but generates it using logic different from that of `GatewayCluster`. The code in `distributed.Client` should be updated in some backwards-compatible way to infer the dashboard url from the cluster object.

txu0393:
Fixed the dashboard URL for the client returned in GatewayCluster.get_client()

distributed.Client.dashboard_link (See https://github.com/dask/distributed/blob/master/distributed/client.py#L750-L764) is implemented in a way such that if the scheduler address has a non-root (i.e. /) path, and/or if the scheduler proxy does not use the same port as the internal scheduler port, the dashboard link will be generated incorrectly. The final dashboard link will have the path dropped, and the port will be set to use the internal port rather than the proxy port .The dashboard link's host is generated correctly.

Is the dashboard URL always going to be the scheduler URL + "/status"?
If that's the case, then the logic for getting distributed.Client.dashboard_link should be updated to reflect that relationship.
We will still need a way to deduce the protocol for the dashboard link.

An alternative fix to this would be to set the "cluster" attribute of a distributed.Client in GatewayCluster.get_client() (by setting it to "self").
This change could be a bit more complicated because the distributed.Client expects "cluster" to be LocalCluster-compatible.

Temporarily resolves #247

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,8 +13,8 @@ pip install \
     ipywidgets \
     jupyterhub \
     notebook \
-    "pytest>=5.4.0" \
-    pytest-asyncio \
+    pytest \
+    pytest-asyncio==0.10.0 \
     sqlalchemy \
     tornado \
     traitlets \
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,7 +13,7 @@ pip install \
     ipywidgets \
     jupyterhub \
     notebook \
-    pytest \
+    "pytest>=5.4.0" \
     pytest-asyncio \
     sqlalchemy \
     tornado \
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,7 +13,7 @@ pip install \
     ipywidgets \
     jupyterhub \
     notebook \
-    "pytest<5.4.0" \
+    pytest \
     pytest-asyncio \
     sqlalchemy \
     tornado \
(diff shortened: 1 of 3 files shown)


txu0393:
Anyone knows what's going on in the CI?
`continuous_integation/install.sh` explicitly  lists requirement `pytest<5.4.0`

gforsyth:
Latest release of `pytest-asyncio` requires `pytest >= 5.4.0`: https://github.com/pytest-dev/pytest-asyncio/blob/master/setup.py#L44

The `pytest` pin was just 15 days ago, and probably to keep the new `pytest` release from breaking `pytest-asyncio` -- now that there's a new `pytest-asyncio` release, we can unpin `pytest` in `continuous_integration/install.sh`

jcrist:
Hi, thanks for the PR. I don't think we want to go this route, but you did bring up a good point with:

> An alternative fix to this would be to set the "cluster" attribute of a distributed.Client in GatewayCluster.get_client() (by setting it to "self").
This change could be a bit more complicated because the distributed.Client expects "cluster" to be LocalCluster-compatible.

The only reason `get_client` exists is as a stop-gap while some weird shutdown behavior with `distributed.Client` is resolved (clients sometimes hang around in the background and spam the proxy for forever, `get_client` lets us explicitly close them when the cluster shuts down). *But* you can also create a client with `Client(cluster)` just fine - in this case `client.cluster` is set, and things do work. I totally missed this.

Could you update your PR to fix `get_client` to set the `cluster` attribute?

txu0393:
Whew, it was a race scenario with the unit test failing (weird)
PR should be good to go now @jcrist!

jcrist:
Thanks @txu0393!

PR #258, opened 2020-04-22
Clarify distinction between scheduler and worker pods

droctothorpe:
Currently, the only way to differentiate the scheduler pods from the worker pods when running `kubectl get pods` is the length of the uid at the end of the pod name.

For example:
```
❯ kgp
NAME                                            READY   STATUS    RESTARTS   AGE
dask-gateway-68h7l                              1/1     Running   0          24h
dask-gateway-6db85d6aa6e9416ca92d91fe520e1b20   1/1     Running   0          23h
```

This is problematic from an administrative standpoint and reduces clarity in our Kubecost Prometheus dashboard.

The change also includes the cluster name in the worker pod name so that we can see, at a glance, which schedulers are tied to which workers.

Also, @jcrist, how would you feel about using `dg` as the prefix instead of `dask-gateway` to reduce the length of the pod name?

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -13,8 +13,8 @@ pip install \
     ipywidgets \
     jupyterhub \
     notebook \
-    "pytest<5.4.0" \
-    pytest-asyncio \
+    pytest \
+    pytest-asyncio==0.10.0 \
     sqlalchemy \
     tornado \
     traitlets \
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -1173,9 +1173,9 @@ class KubeController(KubeBackendAndControllerMixin, Application):
             pod["spec"] = merge_json_objects(pod["spec"], extra_pod_config)
 
         if is_worker:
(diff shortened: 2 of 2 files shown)


jcrist:
Note that we do set a label `app.kubernetes.io/component` to `scheduler` or `worker` for scheduler and worker pods respectively. I'm happy to change how we do things though. Will have to think about abbreviating `dask-gateway-` to `dg-`, we'd still want to make sure the meaning is clear.

droctothorpe:
For sure, the label is helpful programmatically -- this is more of an operational improvement for cluster admins (with the added benefit of improving our Grafana integration).

droctothorpe:
Added Gil's [suggestion](https://github.com/dask/dask-gateway/pull/257#issuecomment-617808178). Hoping it addresses the CI issue.

droctothorpe:
Confirmed that it resolved the CI issue.

droctothorpe:
Hey, @jcrist. Checking in about the status of this PR.

jcrist:
Apologies for the delay @droctothorpe. I've updated this PR to drop the `-gateway` component from all k8s object names. It wasn't necessary, and made the names overly verbose.

droctothorpe:
No worries. Thanks, Jim!

PR #259, opened 2020-04-23
DOC: Clarify where option handlers run.

TomAugspurger:
Clarify that this is run on the server.


TomAugspurger:
Seems that pytest-asyncio 0.11.0 was released recently which is incompatible with the pinned version of pytest. I've pinned pytest-asyncio to 0.10.0.

jcrist:
Thanks @TomAugspurger!

PR #266, opened 2020-05-07
[WIP] Cluster resource limits

### LINKED ISSUE #249: Support resource limits for an individual cluster
Previously we supported configuring limits *per user*. These included:
- Max clusters per user
- Max cores per user
- Max memory per user

These were dropped in the refactor, as most backends have a native way of managing resource (cpu/memory) limits. 

- Kubernetes has [ResourceQuota](https://kubernetes.io/docs/concepts/policy/resource-quotas/)
- YARN, PBS, SLURM, etc... have queues

Supporting them in the gateway also complicated the implementation - most decisions are made at the *cluster level*, but limits on cores/memory were handled at the *user level*.

However, using the backend's native method does impose some restrictions on gateway configuration that may not be desirable for certain admins/users. For example, in Kubernetes, `ResourceQuota` objects are associated with a namespace. If you want to have limits per user, each user would need their own namespace.

I don't think going back to the old model is tenable, as I'd prefer to keep all cluster decisions local to information about a single cluster (rather than global information about multiple clusters). But, we can support some simpler limits natively fairly easily. These would include:

- Max workers per cluster
- Max cores per cluster
- Max memory per cluster

These limits would exist on the `ClusterConfig` object, and could thus be influenced by [user-specific cluster options](https://gateway.dask.org/cluster-options.html). So you could change limits on cluster size per user (or per group, etc...).

jcrist:
This adds resource limits *per cluster*. Currently we support:

- Max cores per cluster
- Max memory per cluster
- Max workers per cluster

At runtime these are normalized into a single `cluster_max_workers`
field, which is used to check incoming scale and adapt requests. If a
request from a user exceeds the limit, it is trimmed to be within
bounds, and a warning is raised in the user's terminal notifying them of
the limit.

Still needs tests and docs.

Fixes #249.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -1,7 +1,17 @@
 import asyncio
 
 import aiohttp
-from traitlets import Instance, Integer, Float, Dict, Union, Unicode, default, validate
+from traitlets import (
+    Instance,
+    Integer,
+    Float,
+    Dict,
+    Union,
+    Unicode,
+    default,
+    validate,
+    observe,
+)
 from traitlets.config import LoggingConfigurable, Configurable
 
 from .. import models
@@ -287,20 +297,6 @@ class ClusterConfig(Configurable):
         config=True,
     )
 
-    @validate("scheduler_memory")
-    def _validate_scheduler_memory(self, proposal):
-        if (
-            self.max_cluster_memory is not None
-            and proposal.value > self.max_cluster_memory
-        ):
-            proposed = format_bytes(proposal.value)
-            limit = format_bytes(self.max_cluster_memory)
-            raise ValueError(
-                f"Scheduler memory request of {proposed} exceeds cluster memory "
-                f"limit of {limit}"
-            )
-        return proposal.value
-
     scheduler_cores = Integer(
         1,
         min=1,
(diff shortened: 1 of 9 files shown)


dask-gateway-server/dask_gateway_server/backends/base.py
```diff
@@ -320,6 +346,68 @@ class ClusterConfig(Configurable):
...
+
+        Set to ``None`` for no limit (default).
+        """,
+        allow_none=True,
+        min=0,
+        config=True,
+    )
+
+    @default("cluster_max_workers")
+    def _default_cluster_max_workers(self):
```
jcrist, line 394: If `cluster_max_workers` is explicitly set, then `cluster_max_cores`/`cluster_max_memory` are ignored, even if those restrictions are stricter. I'm waffling on whether that's the right behavior. Right now I think that's confusing, and we should take the min of the 3.

dask-gateway-server/dask_gateway_server/routes.py
```diff
@@ -226,14 +236,27 @@ async def adapt_cluster(request):
     maximum = msg.get("maximum", None)
     active = msg.get("active", True)
 
+    max_workers = cluster.config.get("cluster_max_workers")
+    resp_msg = None
+    if max_workers is not None:
+        if maximum is None:
+            maximum = max_workers
```
jcrist, line 243: Just realized we'll also want to apply this to `minimum` as well.

jcrist: this whole `adapt` call looks to be missing some input validation period.

dask-gateway-server/dask_gateway_server/backends/base.py
```diff
@@ -320,6 +346,68 @@ class ClusterConfig(Configurable):
...
+        help="""
+        The maximum amount of memory (in bytes) available to this cluster.
+        Allows the following suffixes:
+
+        - K -> Kibibytes
+        - M -> Mebibytes
+        - G -> Gibibytes
+        - T -> Tebibytes
+
+        Set to ``None`` for no limit (default).
```
ericdill, line 360: May want to add "Limits may still be imposed by `cluster_max_cores` and `cluster_max_workers`. Similar for the other two

dask-gateway-server/dask_gateway_server/backends/base.py
ericdill: I'm +1 on taking the min of the 3. That seems more intuitive

PR #271, opened 2020-05-29
Add --no-nanny arg to worker cli

chrisroat:
I use --no-nanny for my dask clusters, and wonder if this is safe for a gateway cluster worker, as well.

One test was failing, but it was also failing prior to this change:
test_db_backend.py::test_idle_timeout

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -6,7 +6,7 @@ from traitlets.config import LoggingConfigurable, Configurable
 
 from .. import models
 from ..options import Options
-from ..traitlets import MemoryLimit, Type, Callable
+from ..traitlets import MemoryLimit, Type, Callable, Command
 from ..utils import awaitable
 
 
@@ -231,13 +231,13 @@ class Backend(LoggingConfigurable):
 class ClusterConfig(Configurable):
     """Base class for holding individual Dask cluster configurations"""
 
-    scheduler_cmd = Unicode(
+    scheduler_cmd = Command(
         "dask-gateway-scheduler",
         help="Shell command to start a dask-gateway scheduler.",
         config=True,
     )
 
-    worker_cmd = Unicode(
+    worker_cmd = Command(
         "dask-gateway-worker",
         help="Shell command to start a dask-gateway worker.",
         config=True,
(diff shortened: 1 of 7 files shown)


jcrist:
This seems fine to me, provided the default is still nanny on. I'm hoping to remove the need for a custom `dask-gateway-worker`/`dask-gateway-scheduler` soon, but my bandwidth for doing this is currently pretty small.

Note that custom args aren't configurable in the current implementation, so there's not an easy way for you to set this. I'd probably make `ClusterConfig.worker_cmd` and `ClusterConfig.scheduler_cmd` accept either a string or a list as a command (we'd want to port over [this traitlet class from JupyterHub](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/traitlets.py#L25-L39)), and then update all the places where we build up commands for submission to expect a list rather than a string for the command.

chrisroat:
Default is still nanny on.

I ran with your suggestion on worker_cmd/scheduler_cmd being "Command" traitlets.  Let me know if it's what you were thinking.

chrisroat:
FWIW, I re-build the server and have been using client+server for a bit now.  It seems to be working.

The travis build failure seems also to occur on my box with master, as well.

chrisroat:
@jcrist Is the --no-nanny in dask-gateway-worker act like it does on dask-worker?  I'm getting a KilledWorker with --no-nanny, and it's not near any limits.  There is nothing the K8s logs, either.  It simply goes from a log about my gsutil command executing, to restarting (K8s doesn't report the container restarting, but the dask-gateway-worker is reissued in the logs).  There is plenty of cpu, memory, and disk space.

chrisroat:
Is there still interest in this?  Would there be any follow-up interest if I were to add a --resources flag, as well?

jcrist:
Hi @chrisroat, yes, I'm still interested in merging this. Apologies for the delay, I've been on hiatus from dask work for the last few weeks. I'll try to get this in within the next week or so. I would accept a PR to add `--resources` as well, but am still hoping to remove the need for the `dask-gateway-worker` cli entirely in the near future.

chrisroat:
Sounds good.  No worries on the delay.

Depending on timing, I can wait on the --resources flag.  Working w/o the
extra dask-gatway-* binaries sounds like a great goal!

jcrist:
I tried to push a few small fixes to the PR branch (usually github has this enabled), but didn't have permission so I pushed a new branch with your commits (and the fixes) and merged those in #286. Thanks for your work, this is now in.

Issue #280, opened 2020-06-24
Monitoring with Prometheus and Grafana


TomAugspurger:
I'm looking into adding monitoring for pangeo, and would like to add a few metrics for Dask Gateway. I'm very new to prometheus and grafana, so apologies for any mistakes. Here are some rambling thoughts as I work through things. https://github.com/TomAugspurger/pangeo-cloud-federation/blob/metrics/gke/values.yaml has some helm configuration
for the pangeo deployment I'm testing on.

I notice a couple of questions immediately:

1. Dask exposes prometheus metrics at `scheduler-address/metrics`, but Dask Gateway's schedulers are
   ephemeral. How do we dynamically notify grafana / prometheus that it should add a datasource at an address?
   Should Dask-Gateway act as a broker here?
2. In my setup I'm using JupyterHub auth. Will all prometheus requests need to go through JupyterHub
   to access the dashboard?

I'll continue to post here as I work through things.

cc @droctothorpe (who mentioned prometheus in https://github.com/dask/dask-gateway/pull/258) and @consideRatio

consideRatio:
1. Prometheus can watch the k8s services and inspect if they have annotations about scraping, if they do, prometheus will scrape metrics from the k8s service exposed pods. So, my main question is if the ephemeral schedulers get a k8s services created for them? If so, can they be annotated?

2. Grafana is what expose dashboards, and it in turn will probe prometheus for data to show in the dashboards. So, grafana user -> grafana web interface -> prometheus as one possible data source for created dashboards in grafana. So, grafana will need to manage a userbase of its own, but you can configure it to work against a OAuth2/OpenID Connect identity provider of your choice. The network connectivity to grafana could be done in various ways, one way is to create an ingress resource that together with the ingress controller provides an entrypoint. A setup I manage have it setup like that with nginx-ingress+cert-manager giving that entry HTTPS as well.

TomAugspurger:
> So, my main question is if the ephemeral schedulers get a k8s services created for them? If so, can they be annotated?

They do indeed. So they would need something like

```yaml

TomAugspurger:
Ah, that might run into https://github.com/dask/dask-gateway/issues/274 when using via helm. I'll try applying it manually first.

TomAugspurger:
Hey, cool, that did it!

[...]

PR #282, opened 2020-06-26
Add worker dashboard

TomAugspurger:
xref https://github.com/dask/dask-gateway/issues/280#issuecomment-650362775

I'm unsure about tests for this...

### DIFF
--- a/dask-gateway/dask_gateway/dask_cli.py
+++ b/dask-gateway/dask_gateway/dask_cli.py
@@ -565,6 +565,12 @@ worker_parser.add_argument("--name", default=None, help="The worker name")
 worker_parser.add_argument(
     "--scheduler-address", default=None, help="The scheduler address"
 )
+worker_parser.add_argument(
+    "--dashboard-address",
+    type=str,
+    default=":0",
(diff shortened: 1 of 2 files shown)


dask-gateway/dask_gateway/dask_cli.py
```diff
@@ -576,6 +583,8 @@ async def start_worker(
     scheduler_address=None,
     local_directory="",
     nanny=True,
+    dashboard=True,
```
TomAugspurger, line 586: The default of `True` here matches the `dask-worker` CLI default. It differs from the `distributed.Worker` default of False. I think following the CLI default is appropriate.

TomAugspurger:
I'm not sure why, but my latest commits aren't showing up here.

jcrist:
github is having issues right now :/

TomAugspurger:
@jcrist all green. Did you have any thoughts on the need for tests here?

jcrist:
Fine by me. I'm hoping to remove the need for the `dask-gateway-worker` cli soon™, don't think we need to add tests for kwarg forwarding that should be compatible with the original cli. Thanks for working on this!

Issue #268, opened 2020-05-11
Add standard mapping option for environment variables.


TomAugspurger:
In our Pangeo deployments, we have a need for setting an environment variable on the workers based on user input. This seems like the kind of thing that would be useful for all backends so I'm raising this here while we'll implement it on our own in Pangeo for now.

1. A widget / traitlet for accepting input from the user in the form of a mapping from `{name: value}`.
2. A standard dask-gateway config for this option for all backends supported by dask-gateway, which is then forwarded to the appropriate location.

The main downside I see with 2 is that this would make two ways to specify environment variables for each backend. The "dask-gateway" way, and the backend-native way. For kubernetes that would be

```yaml
gateway:
  backend:
    worker:
      extraContainerConfig:
        env:
          - name: name
            value: value
```

Is that problematic?

cc @rsignell-usgs and @scottyhq.

jcrist:
> A standard dask-gateway config for this option for all backends supported by dask-gateway, which is then forwarded to the appropriate location.

There is already a standard environment variable config for all backends, on `ClusterConfig.environment`: https://gateway.dask.org/api-server.html#c.ClusterConfig.environment

> A widget / traitlet for accepting input from the user in the form of a mapping from {name: value}

I can see a fancy widget being nice here. For now, you might have users pass in some structured input which is then parsed in the backend as part of your `cluster_options` handler.

TomAugspurger:
Thanks, I somehow missed that `ClusterConfig.environment`.

I'll test out a widget in our pangeo deployments later this week and then make a PR here.

scottyhq:
@jcrist and @TomAugspurger - I was trying this out today because I (and others cc @rsignell-usgs) would like to set environment variables to read data from S3 requester-pays buckets. I understand we can change our cluster config and run `helm upgrade`. But, I was also expecting dask-gateway to pick up config from `distributed`, which doesn't seem to be the case. For example, running a [--preload script](https://docs.dask.org/en/latest/setup/custom-startup.html). 

*Works for LocalCluster*
```python
def get_env(env):
    import os
    return os.environ.get(env)

with dask.config.set({'distributed.worker.preload': ["import os; os.environ['AWS_REQUEST_PAYER'] = 'requester'"]}):
    with LocalCluster(n_workers=2) as cluster, Client(cluster) as client:
        print(client.run(get_env, 'AWS_REQUEST_PAYER'))
```
[...]

PR #283, opened 2020-06-29
test-all

TomAugspurger:

PR #284, opened 2020-07-03
Create dict gateway server option

CrispyCrafter:
Create `Dict` gateway server option such that users can input environment variables via the ipython widget options.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/options.py
+++ b/dask-gateway-server/dask_gateway_server/options.py
@@ -296,3 +296,33 @@ class Select(Field):
 
     def json_type_spec(self):
         return {"type": "select", "options": list(self.options)}
+
+
+@field_doc(
+    description="A dictionary field, allowing users to define cluster config dictionary fields",
+    params="""
+    default : string, optional
+        String to be assigned as dictionary value.    
+    label   : string, optional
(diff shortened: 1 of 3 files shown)


TomAugspurger:
Does this fix https://github.com/dask/dask-gateway/issues/268?

I'm looking into the CI issues at https://github.com/dask/dask-gateway/pull/283, but haven't been able to pin them down yet.

CrispyCrafter:
> Does this fix #268?
> 
> I'm looking into the CI issues at #283, but haven't been able to pin them down yet.

This does address a similar use case as #268
Our use case was to add extra python packages to be installed at runtime via environment variables.
This is working in a live environment now by assigning the output to `ClusterConfig.environment`

Problem is I haven't figured out how to setup multiple entries yet. 
i.e. `ClusterConfig.environment` is not overridden but concatenated

CrispyCrafter:
I'm going to close this PR.
Need to investigate passing in a dictionary such as:
`{'EXTRA_PIP_PACKAGES': 'fbprohet', 'EXTRA_APT_PACKAGES': 'lib-pqdev' }`  that returns a list of Fields.

Optionally this could also be parameterised to specify the Field type, i.e. Number, Bool, Float etc.

PR #290, opened 2020-07-16
Add mapping option

### LINKED ISSUE #268: Add standard mapping option for environment variables.

jcrist:
A cluster options that accepts a mapping. In the ipywidget we parse from yaml, which seems flexible and readable enough. On parse/validation error a red-x appears next to the textarea with a tooltip indicating the error. We also improve the general css of the input form.

Fixes #268, supersedes #284.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/options.py
+++ b/dask-gateway-server/dask_gateway_server/options.py
@@ -1,3 +1,4 @@
+import copy
 import textwrap
 from collections import OrderedDict
 from collections.abc import Sequence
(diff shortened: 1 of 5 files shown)


jcrist:
A few screenshots:

Screen Shot 2020-07-16 at 3 39 36 PM

If invalid yaml is entered, a red `X` appears next to the input form. Hovering over the `X` will display a message about the kind of error.

Screen Shot 2020-07-16 at 3 40 04 PM

Issue #298, opened 2020-08-04
Support TLS for IngressRoute


eyal-lantzman:
When Dask Gateway is hosted in a secure corporate environment (that prohibits accessing the dashboard via http), we still want to provide that ability to the users.
At the minimum, be able to support the follow generated route:
```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
....
spec:
  entryPoints:
  - websecure
  tls:
    options: 
      domains:
      - main: eyal-dask-cpu.812896951795.dev.aws.jpmchase.net
   routes:
     - kind: Rule
...
```

PR #299, opened 2020-08-04
#298: Add support to TLS domains for dashboards

eyal-lantzman:
See related issue: https://github.com/dask/dask-gateway/issues/298

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -9,7 +9,7 @@ import uuid
 from base64 import b64encode
 
 from aiohttp import web
-from traitlets import Unicode, Integer, Float, List, validate
+from traitlets import Unicode, Integer, Float, List, Dict, validate
 from traitlets.config import catch_config_error
 
 from kubernetes_asyncio import client, config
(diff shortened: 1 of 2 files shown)


jcrist:
IIUC the goal here is to have the dashboard traffic served to the user over TLS, is that correct?

If so, I think an easier way would be to add TLS configuration support for both of the existing entrypoints (web and tcp). Then both the api and dashboard traffic could be served over TLS without the need for additional logic in the controller.

I'm not sure how best to offer TLS configuration via the helm chart. Ideally we could support both self-provided certs as well as letsencrypt generated ones. I'd probably look at what the [zero-to-jupyterhub-k8s](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/) helm chart offered for inspiration. Thoughts?

eyal-lantzman:
Good point, I need to provide more context as to why I went for this implementation:
For organizations that can leverage helm charts, that might be a good solution, but in other organizations things like RBAC, restricted access to K8s secrets and special resource-type specific annotations prevent leveraging helm charts from being useful (assuming the organization allows using of helm charts in the first place). 
If you take all those things into consideration, then hard-coded traefik resources in the controller are the real pain point here, the other K8s resource in there have extraXXXConfig which is very useful!

I'm trying to bring dask gateway into the company I'm working for, and helm chart was useful to generate the initial yaml files, which then had to be tweaked quite a bit. Perhaps, once I get it all working I can contribute the additional extension points to helm charts to make it usable in highly controlled K8s environments.

I managed to solve many issues (inc. securing the API) and this PR is one of the few remaining outstanding issues (still investigating some of the other ones). I chose the minimal change to reduce likely hood for breaking changes and nasty side effects. 

If there's a way to configure traefik endpoint with the settings in the PR (not certs/secrets!) and is external to the controller, please let me know, as that would reduce the need for this PR altogether.

eyal-lantzman:
@jcrist , any push back or concerns about this targeted fix?

PR #300, opened 2020-08-05
Remove custom dask-gateway scheduler/worker CLIs

### LINKED ISSUE #177: Support dask-cuda workers
dask-gateway appears to only support starting a cpu-based workers via `dask-gateway-worker`. In order to run [dask-cuda](https://github.com/rapidsai/dask-cuda), dask-gateway needs to be able to invoke `dask-cuda-worker` as well as pass GPU specific variables to `dask-cuda-worker`. Can dask-gateway be extended to support GPUs via RAPIDs? @mmccarty

jcrist:
Previously dask-gateway relied upon custom scheduler/worker CLIs for
integrating with the backend system. While simple to write, this
resulted in a few complications:

- Extra CLI arguments had to be added to support unexposed options
- Users couldn't swap out the dask-worker for a dask-cuda-worker

This PR removes the custom scheduler/worker CLIs in favor of relying on
a preload script, and altering slightly the communication patterns to
avoid the need for anything custom in the workers.

Fixes #177

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -242,15 +242,11 @@ class ClusterConfig(Configurable):
     """Base class for holding individual Dask cluster configurations"""
 
     scheduler_cmd = Command(
-        "dask-gateway-scheduler",
-        help="Shell command to start a dask-gateway scheduler.",
-        config=True,
+        "dask-scheduler", help="Shell command to start a dask scheduler.", config=True,
     )
 
     worker_cmd = Command(
-        "dask-gateway-worker",
-        help="Shell command to start a dask-gateway worker.",
-        config=True,
+        "dask-worker", help="Shell command to start a dask worker.", config=True,
     )
 
     environment = Dict(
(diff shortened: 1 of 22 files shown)


mmccarty:
cc @gforsyth @dankerrigan

dask-gateway-server/dask_gateway_server/backends/base.py
```diff
@@ -242,15 +242,11 @@ class ClusterConfig(Configurable):
...
-        help="Shell command to start a dask-gateway scheduler.",
-        config=True,
+        "dask-scheduler", help="Shell command to start a dask scheduler.", config=True,
     )
 
     worker_cmd = Command(
-        "dask-gateway-worker",
-        help="Shell command to start a dask-gateway worker.",
-        config=True,
+        "dask-worker", help="Shell command to start a dask worker.", config=True,
```
mrocklin, line 249: Woot!

jcrist:
Hmmm, that's a good point. Currently the logic in dask-gateway assumes one process-like-object == one worker. Things will likely get wonky if you spawn multiple processes per CLI command run. We could fix that, would just take some book-keeping. I don't have any dask-cuda experience, so I'm not sure what it does, but the `--nprocs` flag on `dask-worker` would have the same affect. Would you mind opening an issue asking for multiple-processes-per-CLI-call support?

chrisroat:
Done: https://github.com/dask/dask-gateway/issues/319

PR #303, opened 2020-08-11
Pass public_address to creation of new cluster

zonca:
Currently a cluster created by a gateway with a `public_address` different from `address` doesn't inherit it.

Currently I need this workaround:

```
cluster = gateway.new_cluster(public_address = gateway._public_address)
```


jcrist:
Thanks for the PR! Apologies for the delayed response here. Merging.

Issue #304, opened 2020-08-20
helm deployment with `traefik.service.type=ClusterIP` fails


shaunc:
**What happened**:

When I specify "ClusterIP" for the service type I get the error:
```
Unable to deploy helm chart: cannot patch "traefik-dask-gateway" with kind Service: Service "traefik-dask-gateway" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
```

**What you expected to happen**:

Able to deploy with ClusterIP

**Minimal Complete Verifiable Example**:

Should be simple to recreate

**Anything else we need to know?**:

I would like to use dask-gateway with port forwarding for testing, [until I get openid authentication figured out](https://stackoverflow.com/questions/63497961/dask-gateway-helm-deployment-how-to-configure-a-custom-authenticator). In any case I probably would like to deploy behind ingress-nginx and do tls-termination there (with letsencrypt provided by cert-manager).

**Environment**:

- cluster: EKS
- Dask version: dask-gateway 0.8.0 (latest chart)
- Python version:
- Operating System:
- Install method (conda, pip, source): helm
```
helm version
version.BuildInfo{Version:"v3.0.1", GitCommit:"7c22ef9ce89e0ebeb7125ba2ebf7d421f3e82ffa", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

droctothorpe:
Any particular reason why the default service type of `LoadBalancer` doesn't suit your needs? 

On EKS, it should [automatically](https://aws.amazon.com/premiumsupport/knowledge-center/eks-kubernetes-services-cluster/) spin up an AWS load balancer for you.

You can retrieve the ELBS DNS with a simple `kubectl get services` call and just pass that through to the client. 

You can also use [external-dns](https://github.com/kubernetes-sigs/external-dns) to associate a persistent CNAME with the dynamic ELB DNS using a simple annotation on the service.

shaunc:
Mainly security -- I'd prefer to monitor everything going through one LB / ingress rather than add LBs (and authentication protocols).

When we go to replicate this solution, there is a small incremental cost per LB that I'd prefer not to pay.

droctothorpe:
Fair points! Are you planning to interact with the Gateway exclusively from within the cluster? 

> ClusterIP: Exposes the Service on a cluster-internal IP. Choosing this value makes the Service only reachable from within the cluster. This is the default ServiceType.
[...]

PR #305, opened 2020-08-20
add `nodePort: null` to traefik service ports config

shaunc:
Addresses deployment with service type `ClusterIP` -- https://github.com/dask/dask-gateway/issues/304

### DIFF
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -205,10 +205,12 @@ traefik:
       web:
         # The port HTTP(s) requests will be served on
         port: 80
+        nodePort: null
       tcp:
         # The port TCP requests will be served on. Set to `web` to share the
         # web service port
         port: web
+        nodePort: null
 
   # Settings for nodeSelector, affinity, and tolerations for the traefik pods
   nodeSelector: {}


resources/helm/dask-gateway/values.yaml
```diff
@@ -205,10 +205,12 @@ traefik:
       web:
         # The port HTTP(s) requests will be served on
         port: 80
+        nodePort: null
```
jcrist, line 208: Odd, wouldn't the default be `nodePort: null` if it's not present? I'm confused why the existing check in the template results in an invalid spec. Will try to give this a closer look tomorrow morning.

shaunc: Yes -- odd. Perhaps the way the templating system works, missing actually evaluates to an empty string, which then doesn't evaluate to "false". Still - presumably if its referred to as if its in "values.yaml", its good to actually have it there. The other option would be to take it out of "values.yaml" and also out of the service template.

shaunc: (keep in mind that this is golang-based, so that python intuitions might not apply. :))

jcrist:
Apologies for the delay here. I'm unable to replicate your issue locally (setting `traefik.service.type` to `ClusterIP` works fine for me). Not sure why you're having issues (perhaps a bug in an older version of helm?). The changes here aren't harmful though, happy to merge them. Thanks!

PR #309, opened 2020-09-06
feat: add ability to build docker images using python 3.8

shaunc:
Not sure if you were looking for this ability -- but, just in case, this adds:
* docker ARGs to `dask-gateway-base` to switch base image and Miniconda version
* docker ARGs to`dask-gateway` and `dask-gateway-server` to switch base image (to take advantage of above)
* simple script to build `dask-gateway-base` using `python:3.8-slim-buster` base and appropriate Miniconda version for python 3.8 support.

Does not address CI/CD or testing of built images. I did use to build my own base image that does use python3.8 and looks successful (preliminarily).

Passing in of the base could be useful separately. (For instance, I usually use ubuntu for python images.)

### DIFF
--- a/dask-gateway-server/Dockerfile
+++ b/dask-gateway-server/Dockerfile
@@ -1,4 +1,5 @@
-FROM daskgateway/dask-gateway-base:0.7.1 AS dependencies
+ARG base_image=daskgateway/dask-gateway-base:0.7.1
+FROM ${base_image} as dependencies
 LABEL MAINTAINER="Jim Crist-Harif"
 
 # Install dependencies
--- a/dask-gateway/Dockerfile
+++ b/dask-gateway/Dockerfile
@@ -1,4 +1,5 @@
-FROM daskgateway/dask-gateway-base:0.7.1 as dependencies
+ARG base_image=daskgateway/dask-gateway-base:0.7.1
+FROM ${base_image} as dependencies
 LABEL MAINTAINER="Jim Crist-Harif"
 
 # Does the following in one layer
--- a/resources/helm/images/dask-gateway-base/Dockerfile
+++ b/resources/helm/images/dask-gateway-base/Dockerfile
@@ -1,8 +1,16 @@
-FROM debian:10.4-slim
+# python 3.8: -> python:3.8-slim-buster
+ARG base_image=debian:10.4-slim
+FROM ${base_image}
+# FROM python:3.8-slim-buster
 LABEL MAINTAINER="Jim Crist-Harif"
 
-ENV CONDA_VERSION py37_4.8.3
-ENV CONDA_SHA256 bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a
+# python 3.8: -> py38_4.8.3
+ARG conda_version=py37_4.8.3
+# pythom 3.8" -> 879457af6a0bf5b34b48c12de31d4df0ee2f06a8e68768e5758c3293b2daf688
+ARG conda_sha256=bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a
+ENV CONDA_VERSION ${conda_version}
+ENV CONDA_SHA256 ${conda_sha256}
+ENV DEBIAN_FRONTEND noninteractive
 
 # We do the following all in one block:
 # - Create user dask
--- a/resources/helm/images/dask-gateway-base/build_python3.8.sh
+++ b/resources/helm/images/dask-gateway-base/build_python3.8.sh
@@ -1,0 +1,5 @@
+# Build docker image with arguments appropriate for python3.8
+#
+# $1 should be tag for image
+tag=${1:=daskgateway/dask-gateway:0.7.1-py38}
+docker build --build-arg base_image=python:3.8-slim-buster --build-arg conda_version=py38_4.8.3 --build-arg conda_sha256=879457af6a0bf5b34b48c12de31d4df0ee2f06a8e68768e5758c3293b2daf688  -t $tag .


tanishq-dubey:
Was this tested on something like K8S?

shaunc:
I have it running in k8 -- but "works for me" isn't necessarily definitive.

tanishq-dubey:
Better than nothing! Do you have a config that you would be willing to share? I've been operating off the default helm chart, and simply changing to the py3.8 image doesn't work as the `dask_gateway` cli import was removed

shaunc:
Well -- sort of messy right now ... I'm using the helm chart from this PR... and Dockerfile to build worker image:
```dockerfile
# syntax=docker/dockerfile:experimental

FROM reg.factfiber.ai/factfiber/antegraph/dask-gateway:0.8.0-py38 as base

# git+ssh in requirements.txt requires git
USER root
RUN apt-get update && apt-get install -y git
USER 1000:1000

# install extra python packages
COPY requirements.txt .
COPY --chown=1000:1000 ssh/known_hosts /home/dask/.ssh/known_hosts
RUN chmod 0700 /home/dask/.ssh && chmod 0600 /home/dask/.ssh/known_hosts
RUN --mount=type=ssh,id=gitlab,uid=1000,gid=1000  pip install --user -r ./requirements.txt

FROM reg.factfiber.ai/factfiber/antegraph/dask-gateway:0.8.0-py38 as prod
COPY --from=base --chown=1000:1000 /home/dask/.local /home/dask/.local
ENV PATH=/home/dask/.local/bin:$PATH

```
... which you'll want to amend for yourself... and then chart values:
```yaml
      traefik:
        loglevel: DEBUG
        dashboard: true
        service:
          type: ClusterIP
          ports:
            web:
              nodePort: null
              port: 80
            tcp:
              nodePort: null
              port: web
      controller:
        image:
          name: "reg.factfiber.ai/factfiber/antegraph/dask-gateway-server"
          tag: 0.8.0-py38
        imagePullSecrets:
        - name: devspace-auth-reg-factfiber-ai

      gateway:
        image:
          name: "reg.factfiber.ai/factfiber/antegraph/dask-gateway-server"
          tag: 0.8.0-py38
        imagePullSecrets:
        - name: devspace-auth-reg-factfiber-ai

        backend:
          image:
            name: "reg.factfiber.ai/factfiber/antegraph/dask-worker"
            tag: ${DASK_WORKER_VERSION}
          scheduler:
            extraPodConfig:
              imagePullSecrets:
              - name: devspace-auth-reg-factfiber-ai
          worker:
            extraPodConfig:
              imagePullSecrets:
              - name: devspace-auth-reg-factfiber-ai
        extraConfig:
```
note that I pushed the images to my own repository from which I'm pulling. Also I'm using devspace -- which interpolates `${DASK_WORKER_VERSION}`. If you start from here, will require a bit of tinkering! ;)

tanishq-dubey:
Thanks for the help! I'll try some of these things out and post my results as well, perhaps they can help with an eventual merge!

jcrist:
Apologies for the delayed response here. I'm a bit hesitant to merge this. We see the provided images as examples, but if a user wants to do something more complicated/custom I'd rather they either:

- Extend our provided ones directly by using them as a base image
- Create their own image from scratch. We document the requirements, but we could expand this if needed to be more descriptive. What would be useful to see here?

I don't really see the dockerfiles we publish as a public interface. The internals of our buildsystem shouldn't really leak out to users. JupyterHub has the same policy, the provided images are just examples.

tanishq-dubey:
Sure, that is understandable. I think the point of the merge is more to _at least_ document how someone can run dask-gateway with py3.8 and if there is a path forward on creating official documentation or code allowing for a merge, I'd be more than happy to help with that.

shaunc:
@jcrist IMO the build system needs to extend to different python versions -- the PR is sort of a stab in the dark in the direction of that goal, and I'm sure it could be done in a much more principled manner combined with CI/CD. The script and the Dockerfiles arguments I thought might be useful for such a system, even if they don't constitute a full solution. They also allow for simpler customization/extension even if that isn't directly supported.

jcrist:
There isn't anything special about python 3.8 that would be different than python 3.7. The issue you're seeing above:

> Do you have a config that you would be willing to share? I've been operating off the default helm chart, and simply changing to the py3.8 image doesn't work as the dask_gateway cli import was removed

is because you're mixing the master branch here and the latest release. The CLIs were removed after the last release, but a new release hasn't been made yet. Using the published chart at https://dask.org/dask-gateway-helm-repo/ with the latest release of dask-gateway (0.8.0) from pypi/conda-forge works fine with any supported version of python.

jcrist:
> IMO it is desirable that the build system needs to extend(s) to different python versions

Again, I don't see our `Dockerfile`s as public interfaces. The released images are public and can be used, but I'd rather users that need customization develop their own image than use ours with a bunch of extra build args. I'd be happy to add whatever documentation or examples are needed to make this easier for users, but I'm -1 on adding anything that makes our Dockerfile itself a public interface.

shaunc:
Ok ... that makes sense.... On the one hand I wanted to build custom images in the simplest possible manner, and on the other I thought the project itself would want an easy way to role forward to other python versions. I was hoping to finesse the two goals. It doesn't necessarily mean that by accepting this you make the Dockerfiles into public images. Just you make them easier to use for a hacker who happens to come across them. :) 

Still, the putative rationale for the project would just be able to support various python versions, not to make the Dockerfiles easily extendable by others.

jcrist:
The existing dockerfiles have been quite easy to maintain, adding extra switches (and worrying that people might be using them, meaning I can't change them) adds some maintenance cost.

---

I'm thinking about adding an example `dask-gateway` Dockerfile that we could use to document how users might create their own images. This would live in the `resources/helm/images` folder, and would never be built by us directly (example only). What base image is the easiest way to get conda now? Anaconda used to publish a base image, but that's quite out of date now.

Perhaps a dockerfile that was:

- based on some common base image like debian or ubuntu
- installed conda in a simple one-line run command
- then installed dask-gateway

would be sufficient.

A lot of the complications of these docker images come from:

- Building dask-gateway from source. A user can always install it in the `Dockerfile` from conda-forge or PyPI.
- Stripping excess files. This shouldn't be needed for most users.

I'd like to write one that documents the bare minimum requirements (a valid `dask-gateway` install) in a readable way. Thoughts?

shaunc:
I guess maybe what you suggest is sufficient. That image could always be updated as necessary and even people who are rolling their own could use as a reference.

I like to use the "official" python images built on "slim". I also like to try to strip extra files, but perhaps that is over optimization. As for Conda -- [you seem to be the expert in the field](https://jcristharif.com/conda-docker-tips.html). :). Getting the various c libraries installed using pip (+apt-get & environment settings, switches, etc) without conda would certainly be more difficult than installing Conda however its done. Maybe someone needs to create a new "python-x.y-slim-conda" image to use as the base of the base....

jcrist:
In #321 we restructured the docker images to be more straightforward to read for curious users. We also added an example dockerfile for developing your own `dask-gateway` images here: https://github.com/dask/dask-gateway/blob/master/resources/helm/example-images/Dockerfile. A section has been added to the documentation as well on the requirements for custom images, with a link to the example dockerfile (will be [here](https://gateway.dask.org/install-kube.html#additional-configuration) eventually once the build updates).

Please let me know if there's anything else that would be useful for you here.

PR #310, opened 2020-09-07
feat: add imagePullSecrets option to values for gateway and controller

shaunc:
As an alternative to adding "imagePullSecrets", one could have "extraPodConfig" as you do with worker and scheduler. That seemed to me like too large a change in the templates. 

Putatively, I could use an external service account instead for pull secrets, but then would have to track the needed role bindings if they changed. In theory, imagePullSecrets option could be added to the default service account config as well or instead.

### DIFF
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -22,6 +22,9 @@ gateway:
     tag: 0.8.0
     pullPolicy: IfNotPresent
 
+  # Image pull secrets for gateway-server pod
+  imagePullSecrets: []
+
   # Configuration for the gateway-server service
   service:
     annotations: {}
@@ -141,6 +144,9 @@ controller:
 
   # Resource requests/limits for the controller pod
   resources: {}
+
+  # Image pull secrets for controller pod
+  imagePullSecrets: []
 
   # The controller log level
   loglevel: INFO
--- a/resources/helm/dask-gateway/templates/controller/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/controller/deployment.yaml
@@ -35,6 +35,10 @@ spec:
         - name: configmap
           configMap:
             name: {{ include "dask-gateway.controllerName" . }}
+      {{- with .Values.controller.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: controller
           image: {{ .Values.controller.image.name }}:{{ .Values.controller.image.tag }}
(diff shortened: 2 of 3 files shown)


resources/helm/dask-gateway/values.yaml
```diff
@@ -22,6 +22,9 @@ gateway:
     tag: 0.8.0
     pullPolicy: IfNotPresent
 
+  # Image pull secrets for gateway-server pod
+  imagePullSecrets: []
```
jcrist, line 26: Would you mind moving this under `image` as `image.pullSecrets`? Same for the other locations.

shaunc: Hmm... so on my branch I'm experimenting with configuration to allow one to inject sidecar containers into pods. Pretty simple from the helm perspective -- just a blank space. I'm using it for an auth proxy, but I would hope there are other uses for it as well.

Given this ... `imagePullSecrets` are part of the pod spec, not the container spec. In some sense it wouldn't matter as they would have to be put into the pod spec anyway, and so would work. But logically it would be a bit of a hack when you had multiple containers to specify the `imagePullSecrets` as part of the image specification for the "main" container, while they might actually only be needed for a sidecar.

Of course -- you might be ok with this. You may tolerate my sidecar patch (or not), but putatively it's not a central use case for you. I'll move it if you want but just wanted to explain my hesitation.

droctothorpe: Just out of curiosity, @shaunc, why not put the auth proxy in a standalone pod as opposed to a sidecar?

shaunc: This is best practice to deploy [louketo proxy](https://github.com/louketo/louketo-proxy). The idea is that you don't have to publish the non-auth port at all.

In fact, I don't actually need this right now. Rather, I have a few different applications that I've setup like this, and out of laziness I'm hoping to cut & paste the setup. Careful RBAC setup would also allow us to achieve the same goals. (Another option would be istio/envoy, but I've tried it twice and ripped it out both times ... not mature enough and/or too complex!)

But that said, in the past we have worked with data that requires strict HIPPA compliance. To get the cluster into compliance will require going through an enormous checklist. If it becomes necessary having the same secure setup in many places will make the job easier.

jcrist: Apologies for the delayed response here. You're adding a sidecar container to the controller/gateway pods? How are you adding that config via helm?

Since the common case is adding pull secrets for the pods/containers that are already part of the helm chart, I'd still prefer if the pull secrets were under the corresponding `image` keys. This also matches the convention in several other helm charts we make use of, see e.g. JupyterHub:

https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/26eb298b19f472b35d7a5267036918e1252941cd/jupyterhub/values.yaml#L48-L51

jcrist: After further consideration, I think this is fine as is. Merging.

PR #321, opened 2020-10-02
Restructure docker images

jcrist:
- Remove the `dask-gateway-base` image. Having a base image
  over-complicated things, it wasn't clear when to update the base
  image, and most docker tooling doesn't play as nice with images built
  from multiple Dockerfiles. We fix the problem by avoiding the problem.
  The `dask-gateway-server` image now is based on one of the official
  python images (currently python 3.8), and the `dask-gateway` image
  installs miniconda as part of its standard build.

- Add an example `Dockerfile` users can use to develop their own images.
  Several users have struggled with building valid images, probably
  because they're using our build-system `Dockerfiles` as a reference.
  These do some extra work to build dask-gateway from source, which a
  user shouldn't need to do. We add an example installing dask-gateway
  from conda-forge, with a heavily commented dockerfile that users can
  use as a reference.

Supersedes/fixes #309.


jcrist:
Hmmm, the build is timing out, but I'm unable to reproduce locally (tests pass fine locally). I'd like to reorg the k8s tests to get actual logs from the services, so I'm going to ignore this failure for now (again, things pass fine locally).

jcrist:
Alright, I've at least validated that things run fine locally. Not sure why the travis tests are failing (and without restructuring our tests/switching off travis ci, it's hard to get logs as to why. See #322.). Merging.

shaunc:
Nice -- well you remove some of the optionality but generally leave me little reason to need it. :) I didn't know about tini.
[NB -- should note I just hit https://github.com/pandas-dev/pandas/issues/35917 with pandas 1.1.2, so of course always room for quibbles... but doc clearly gives example on how to build.

Thanks!

PR #325, opened 2020-10-05
Add Gateway and Controller affinities, nodeSelector, tolerations

TomAugspurger:
This PR allows setting the affinities, nodeSelector, and tolerations for the Gateway and Controller pods. I may be wrong, but I didn't see a way to do that from the `extraConfig`.

It also adds a very basic test, ensuring that the helm chart can be deployed with `--dry-run`. That would have caught the other helm issue. I can remove it if we don't think it's useful (or expand it to cover other things if you want). Apparently helm has some tools for testing charts, though I've never used them before.


jcrist:
Thanks Tom, this looks good to me. Merging.

jcrist:
> Apparently helm has some tools for testing charts, though I've never used them before.

We use `helm lint` and test the validity of the generated yaml as part of the `pytest` tests. I'm hoping to debug and fix the travis kube test setup this week, when I do that I'll port over your `dry-run` test (and add a few other configurations to test) as well.

PR #326, opened 2020-10-07
Try out CircleCI

jcrist:
CircleCI supports running multiple containers with live-updating logs, which may make it easier to debug CI failures for dask-gateway, which relies on multiple services in its tests.


jcrist:
I was able to get the features I wanted (logs from a background service) without requiring switching CI services. CircleCI has some nice features, but they're not worth porting over this existing project. Closing.

PR #334, opened 2020-10-13
Add LSF scheduler backend

### LINKED ISSUE #333: Support LSF scheduler
It would be great if we could support the IBM LSF scheduler for running, e.g., on Summit.

To that end, I already implemented it in a branch and will create a PR.

jglaser:
Implement an LSF scheduler backend. To make it work on the system of interest (Summit at ORNL), I had to make the scheduler and API ports configurable due to firewall restrictions.

resolves #333 

- [ ] update docs
- [ ] write unit tests

### DIFF
--- a/dask-gateway-server/dask-gateway-proxy/fail_without_go_1_12.go
+++ b/dask-gateway-server/dask-gateway-proxy/fail_without_go_1_12.go
@@ -1,7 +1,7 @@
-// +build !go1.12
+// +build !go1.11
 
 package main
 
 func compileTimeError() {
-	GO_1_12_REQUIRED_FOR_BUILD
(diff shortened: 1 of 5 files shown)


consideRatio:
@jglaser I'm sorry for your experience of not getting any feedback about this and #333 in the past!

The maintenance capacity for this project is low and my aim is to keep it alive and functional but enter a sustainable maintenance phase where we avoid growing it unless maintenance capacity increase and such. For new backends, I suggest they are declared in a separate repo and maintained by the author at this point.

For reference for anyone reading this, @jglaser work on a LSF scheduler backend is found at https://github.com/jglaser/dask-gateway/tree/lsf_backend still.

PR #336, opened 2020-10-13
Add extension mechanism for side-loading config

droctothorpe:
This PR provides a mechanism for adding configuration / extensions without needing to install them through the Dockerfile (which requires managing your own image) or interleaving your Python code with Helm yamls via the `extraConfig` field in `values.yaml`. **This makes it possible to implement unit tests and continuous integration for extensions.**

To use this method, simply copy your extensions as either packages or modules
into the extensions directory or sub-directories within it.

Helm will copy the contents of any files that end in `_config.py` into
`dask_gateway_config.py`.

I have validated that this PR works with our auth and options extensions after appending the relevant files with `_config.py`.

Happy to submit a documentation PR as well if this gets approved.

This PR is fully backwards-compatible. It provides an alternative, not a replacement.

### DIFF
--- a/resources/helm/dask-gateway/templates/gateway/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/configmap.yaml
@@ -18,6 +18,7 @@ data:
     {{- end }}
     import os
     import json
+    import sys
 
     _PROPERTIES = json.loads({{- $values | toJson | quote }})
 
@@ -107,6 +108,11 @@ data:
     else:
         raise ValueError("Unknown authenticator type %r" % auth_type)
 
+    # Add /etc/dask-gateway/ to the Python path.
+    # This is required for side-loaded extensions to work.
+    configuration_directory = os.path.dirname(os.path.realpath(__file__))
+    sys.path.insert(0, configuration_directory)
+
     {{- if $extraConfig }}
     {{ if kindIs "string" $extraConfig }}
     # From gateway.extraConfig
@@ -119,7 +125,7 @@ data:
     {{- end }}
     {{- end }}
 
-    {{/* Configuration pulled in through *_config.py files in the extensions directory */}}
-    {{ range $path, $_ :=  .Files.Glob  "**_config.py" }}
-        {{ $.Files.Get $path | nindent 4 }}
-    {{ end }}
+  {{- /* Glob files to be added to the ConfigMap. */ -}}
+  {{- /* These files are mounted alongside dask_gateway_config.py */ -}}
+  {{- /* key = filename; value = content of file */ -}}
+  {{- (.Files.Glob "extensions/gateway/*").AsConfig | nindent 2 }}
--- a/resources/helm/dask-gateway/templates/gateway/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/configmap.yaml
@@ -118,3 +118,8 @@ data:
     {{- end }}
     {{- end }}
     {{- end }}
+
+    {{/* Configuration pulled in through *_config.py files in the extensions directory */}}
+    {{ range $path, $_ :=  .Files.Glob  "**_config.py" }}
+        {{ $.Files.Get $path | nindent 4 }}
+    {{ end }}
--- a/resources/helm/dask-gateway/templates/gateway/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/configmap.yaml
@@ -6,9 +6,9 @@ metadata:
     {{- include "dask-gateway.labels" . | nindent 4 }}
 data:
   dask_gateway_config.py: |-
-    {{/* Extract extraConfig, as its handled separately */}}
+    {{- /* Extract extraConfig, as its handled separately */}}
     {{- $extraConfig := .Values.gateway.extraConfig }}
-    {{/* Remove secrets and extraConfig before forwarding configuration */}}
+    {{- /* Remove secrets and extraConfig before forwarding configuration */}}
     {{- $values := pick .Values "gateway" }}
     {{- $_ := set $values "gateway" (omit $values.gateway "extraConfig") }}
     {{- if $values.gateway.auth.jupyterhub.apiToken }}
@@ -19,6 +19,9 @@ data:
     import os
     import json
     import sys
+
+    # Add this directory to path to support extension modules
+    sys.path.insert(0, os.path.dirname(os.path.realpath(__file__)))
 
     _PROPERTIES = json.loads({{- $values | toJson | quote }})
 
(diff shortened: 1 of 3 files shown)


resources/helm/dask-gateway/templates/gateway/configmap.yaml
```diff
@@ -118,3 +118,8 @@ data:
     {{- end }}
     {{- end }}
     {{- end }}
+
+    {{/* Configuration pulled in through *_config.py files in the extensions directory */}}
+    {{ range $path, $_ :=  .Files.Glob  "**_config.py" }}
```
jcrist, line 123: This globs every directory, we'd probably want to limit to `extensions`? Why not drop the `*_config.py` pattern and add a subdirectory of `extensions` only for extra files to add to the `dask-gateway` pod. Something like:

```suggestion
    {{ range $path, $_ :=  .Files.Glob  "extensions/gateway/*.py" }}
```

JupyterHub follows a similar pattern: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/hub/configmap.yaml#L10

jcrist (commented):
Hmmm, interesting. IIUC this requires that users have a local copy of the helm chart, rather than using the published one in the helm repository?

I'm a bit wary of inlining multiple files into the `dask_gateway_config.py` - a folder of python files looks a lot like a python module, I'm afraid users may try to import functionality from one file into another, which would fail in practice. Global variables and imports might also collide, leading to issues.

Instead of inlining the files in the `dask_gateway_config.py` entry in the configmap, we could use `AsConfig` to add configmap entries for each file placing them in the same directory as the `dask_gateway_config.py` file (we'd need to drop the subdirectories support in `extensions`, as only files would work). You could then have the bulk of your code defining new classes or functions in files in `extensions`, and use `extraConfig` to import the bits you need to pass to dask-gateway's configuration.

The `AsConfig` helm helper (https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions) would help here.

resources/helm/dask-gateway/templates/gateway/configmap.yaml
droctothorpe: Thanks for responding so quickly, Jim.

> Hmmm, interesting. IIUC this requires that users have a local copy of the helm chart, rather than using the published one in the helm repository?

This approach does assume a local copy of the Helm chart. 

> 
> I'm a bit wary of inlining multiple files into the `dask_gateway_config.py` - a folder of python files looks a lot like a python module, I'm afraid users may try to import functionality from one file into another, which would fail in practice. Global variables and imports might also collide, leading to issues.
> 
> Instead of inlining the files in the `dask_gateway_config.py` entry in the configmap, we could use `AsConfig` to add configmap entries for each file placing them in the same directory as the `dask_gateway_config.py` file (we'd need to drop the subdirectories support in `extensions`, as only files would work). You could then have the bulk of your code defining new classes or functions in files in `extensions`, and use `extraConfig` to import the bits you need to pass to dask-gateway's configuration.

Correct me if I'm wrong, but won't this require adding [extraVoleMounts](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/6e5ec7e50a8ed670bae4c5d195bb01a82b157a4b/jupyterhub/templates/hub/deployment.yaml#L122) to the Dask Gateway API deployment and the [values.yaml](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/values.yaml#L55) files?

I want to mention two minor drawbacks. First, it requires updates in two locations: the `extensions` directory and `extraConfig` in `values.yaml`. Second, the extensions cannot be structured like packages, which is less of an issue for `options`, but authentication extensions can get fairly complex. Neither of these are deal breakers, and the alternative, naming collisions, is a significant concern -- I just think they're worth mentioning.

I'm happy to update this PR to align with the JupyterHub design pattern if you think it's a better model. 

> 
> The `AsConfig` helm helper (https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions) would help here.

This is a bit of a tangent, and probably deserves a discrete PR, but our authentication extension imports the [pyjwt](https://pyjwt.readthedocs.io/en/latest/) package. I'm wondering if there's a way to side-load that dependency without needing to build our own image or if that's a fool's errand. I suspect other users writing custom authentication modules will run into a similar obstacle.

droctothorpe:
Thanks for responding so quickly, Jim.

> Hmmm, interesting. IIUC this requires that users have a local copy of the helm chart, rather than using the published one in the helm repository?

This approach does assume a local copy of the Helm chart.

> I'm a bit wary of inlining multiple files into the dask_gateway_config.py - a folder of python files looks a lot like a python module, I'm afraid users may try to import functionality from one file into another, which would fail in practice. Global variables and imports might also collide, leading to issues.

> Instead of inlining the files in the dask_gateway_config.py entry in the configmap, we could use AsConfig to add configmap entries for each file placing them in the same directory as the dask_gateway_config.py file (we'd need to drop the subdirectories support in extensions, as only files would work). You could then have the bulk of your code defining new classes or functions in files in extensions, and use extraConfig to import the bits you need to pass to dask-gateway's configuration.

Correct me if I'm wrong, but won't this require adding extraVoleMounts to the Dask Gateway API deployment and the values.yaml files?

I want to mention two minor drawbacks. First, it requires updates in two locations: the extensions directory and extraConfig in values.yaml. Second, the extensions cannot be structured like packages, which is less of an issue for options, but authentication extensions can get fairly complex. Neither of these are deal breakers, and the alternative, naming collisions, is a significant concern -- I just think they're worth mentioning.

I'm happy to update this PR to align with the JupyterHub design pattern if you think it's a better model.

This is a bit of a tangent, and probably deserves a discrete PR, but our authentication extension imports the pyjwt package. I'm wondering if there's a way to side-load that dependency without needing to build our own image or if that's a fool's errand. I suspect other users writing custom authentication modules will run into a similar obstacle.

jcrist:
> Correct me if I'm wrong, but won't this require adding extraVolumeMounts to the Dask Gateway API deployment and the values.yaml files?

This isn't correct. We'd be storing all the "extensions" in a single configmap, but they'd be mounted as separate files rather than concatenated.

> Second, the extensions cannot be structured like packages, which is less of an issue for options, but authentication extensions can get fairly complex.

They can't be structured like packages in your implementation either, as imports between modules won't work. I think that making them look like packages without supporting things like imports or name collisions is likely to result in user confusion.

I agree that two places to configure something  would be unfortunate, but I think the complications of concatenating python files outweighs the benefits.

> This is a bit of a tangent, and probably deserves a discrete PR, but our authentication extension imports the pyjwt package. I'm wondering if there's a way to side-load that dependency without needing to build our own image or if that's a fool's errand. I suspect other users writing custom authentication modules will run into a similar obstacle.

Yeah, I think this is a fools errand. Any extra dependencies you need you'll need to install in your own images.

droctothorpe:
Hey @jcrist. Thanks for the constructive feedback. You were spot on about the advantages of using discrete modules in conjunction with `AsConfig`. I pushed the suggested changes. Standing by for further input.

jcrist:
Thanks @droctothorpe. I pushed a few fixups, this looks good to me.

droctothorpe:
Awesome. Thanks, @jcrist!

PR #338, opened 2020-10-14
Multiple workers per job

jglaser:
This PR allows multiple worker instances to be assigned to a single job, extending the 1:1 mapping of jobs and workers to N:1. At the low-level, this is enabled by requiring the worker identifiers to have a common prefix and to end with a worker specific identifier after the dash, such as `-0`, `-1`, etc. This is supported through the standard `--nprocs` option in `dask-worker`, and through `dask-mpi` (https://github.com/dask/dask-mpi/pull/52).

A semantic change going along with this is that the `GatewayCluster` interface now allows requesting worker **jobs**, but displays the up-to-date number of **workers**.

[ ] Document
[ ] Unit tests

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -6,6 +6,7 @@ from traitlets import (
     Integer,
     Float,
     Dict,
+    List,
     Union,
     Unicode,
     default,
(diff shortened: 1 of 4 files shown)


consideRatio:
I think this may be a quite relevant PR even still, but the change must be motivated a bit clearer for my ability to review and consider if this is a change to be accepted and maintained long term.

Given the low maintenance capacity I'll go for a close on this at this point. I'm sorry for the lack of input you got @jglaser, I think a good practice to ensure effort isn't wasted in open source contributions is to try establish agreement to implement a feature ahead of time - which is part of the review of a PR anyhow.

Issue #296, opened 2020-07-24
Automate release process


jcrist:
Currently the release process is manual, and has several error-prone steps. It would be good to automate much of this, both for ease of maintenance and to prevent mistakes. The artifacts from a release are:

- `dask-gateway` and `dask-gateway-server` packages on pypi and conda-forge
- `daskgateway/dask-gateway` and `daskgateway/dask-gateway-server` images on docker hub
- A new helm chart packaged to https://dask.org/dask-gateway-helm-repo/

### Releasing the packages

- Update the versions for both `dask-gateway` and `dask-gateway-server`. These are currently stored in `_version.py` files in each package.

- Commit and tag the release

```
git commit -m "Release $VERSION"
git tag -a $VERSION -m "Release $VERSION"
```

- Build `dask-gateway` as an `sdist` and a `bdist_wheel`, and upload to pypi:

```
cd dask-gateway
python setup.py sdist
python setup.py bdist_wheel
twine upload dist/*
```

- Build `dask-gateway-server` as an `sdist` and a `bdist_wheel`, and upload to pypi. Since this package has a binary component, this is a bit more involved:

```
cd dask-gateway-server
python setup.py sdist
python build_tools/build_macos.py
python build_tools.build_manylinux.py
twine upload dist/*
```

- Update the dask-gateway-feedstock on conda-forge, and merge. Usually the conda-forge automation takes care of this for us.

### Releasing the helm chart and docker images

- Build the `dask-gateway` and `dask-gateway-server` docker images and upload to docker hub. If it's been a bit between releases and miniconda has updated a lot, we may also need to build the `dask-gateway-base` image that's shared between those, updating miniconda to a newer release (that dockerfile lives [here](https://github.com/dask/dask-gateway/blob/master/resources/helm/images/dask-gateway-base/Dockerfile)). Generally when releasing I also bump the dependencies in the provided docker images to the latest stable version. We advise users to build their own `dask-gateway` images (since those run user code), but the `dask-gateway-server` image should work for all users and should be thoroughly tested. There's almost certainly a better system than this.

```
docker build dask-gateway-server/ --tag daskgateway/dask-gateway-server:$VERSION
docker tag dask-gateway-server:$VERSION daskgateway/dask-gateway-server:latest
docker build dask-gateway/ --tag daskgateway/dask-gateway:$VERSION
docker tag dask-gateway:$VERSION daskgateway/dask-gateway:latest
docker push daskgateway/dask-gateway-server:$VERSION
[...]

PR #339, opened 2020-10-16
Test GitHub Actions for auto-release

fanshi118:
I was testing the auto-release features using GitHub Actions, taken from [this issue](https://github.com/dask/dask-gateway/issues/296) . Building `dask-gateway` went okay; however, I ran into an error while running `python build_tools/build_manylinux.py` in building `dask-gateway-server`. There seemed to be an issue with running the docker command in [the script](https://github.com/dask/dask-gateway/blob/master/dask-gateway-server/build_tools/build_manylinux.py) . Would be awesome to get some feedback on this!

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -4,10 +4,9 @@ on:
     branches:
       - master
       - auto-release
+    tags: ["*"]
   pull_request:
     branches: [master]
-  release:
-    types: [published]
 
 jobs:
   build_gateway:
@@ -75,7 +74,7 @@ jobs:
     name: Upload to PyPI on release - dask-gateway
     needs: [build_gateway]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'published'
+    if: github.tag != ''
     steps:
       - uses: actions/download-artifact@v2
         with:
@@ -90,7 +89,7 @@ jobs:
     name: Upload to PyPI on release - dask-gateway-server
     needs: [build_gateway_server]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'published'
+    if: github.tag != ''
     steps:
       - uses: actions/download-artifact@v2
         with:
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,0 +1,102 @@
+name: PyPI
+on:
+  push:
+    branches:
+      - master
+      - auto-release
+  pull_request:
+    branches: [master]
+  release:
+    types: [published]
+
+jobs:
+  build_gateway:
+    name: Build source distribution - dask-gateway
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - name: Build dask-gateway as an sdist and a bdist_wheel
+        run: |
+          cd dask-gateway
+          python setup.py sdist
+          pip install wheel
+          python setup.py bdist_wheel
+          cd ..
+      - name: Check the sdist installs and imports
+        run: |
+          mkdir -p test-sdist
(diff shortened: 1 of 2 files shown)


fanshi118:
As an update to the above, I was able to run `build_tools/build_manylinux.py` in GitHub Actions and build `dask-gateway-server` by removing the `-it` at line 45 in the script. @jcrist : think this may have resolved the first half of #296

dgerlanc:
@jcrist Any way we can help to get this merged in? Would be great to have a new release with all recent updates.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,102 @@
+name: PyPI
+on:
+  push:
+    branches:
+      - master
+      - auto-release
+  pull_request:
+    branches: [master]
+  release:
+    types: [published]
```
jacobtomlinson, line 10: Generally with Dask projects we do not use the GitHub releases feature. We just create new git tags and push them.

So I'm not sure this would ever be triggered. We should instead trigger on tags.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,102 @@
+name: PyPI
+on:
+  push:
+    branches:
+      - master
+      - auto-release
+  pull_request:
+    branches: [master]
```
jacobtomlinson, line 8: My understanding is that this would be triggered when a PR is made from master. Is that what we want?

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,102 @@
+name: PyPI
+on:
+  push:
+    branches:
+      - master
+      - auto-release
```
jacobtomlinson, line 6: Do we want this to run on every merge to master? And what is the `auto-release` branch?

jacobtomlinson (commented):
A few comments

.github/workflows/pypi.yml
fanshi118: `auto-release` is a branch I made in my fork. I was testing the feasibility of this auto-release workflow, so made it run on every merge to master.

.github/workflows/pypi.yml
fanshi118: Luckily there's a way to [trigger a GitHub action workflow run on tags, on either push or pull request](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags). What versions are we thinking?

jacobtomlinson: All. You should be able to do something like 

```yaml
on:
  push:
    tags:        
      - *
```

.github/workflows/pypi.yml
jacobtomlinson: Ok sure. That will need removing before merging though.

fanshi118:
@jacobtomlinson - Actually there are two parts of this workflow. One is to build the artifacts and upload them to GitHub (like [here](https://github.com/fanshi118/dask-gateway/actions/runs/350238389)), and the other to upload the package to PyPI. The current way we set it up is to run the first part for every merge to master - that is, rebuild the artifacts for every change made to the main/production branch. The second part, upload to PyPI, will only be triggered if a new tag is pushed. Trying to tie this back to the [original issue](https://github.com/dask/dask-gateway/issues/296) , do we not want the artifacts for this and only push releases for every new tag?

jacobtomlinson:
@fanshi118 that makes sense. I can't comment on whether artifacts are desired on every commit to master. Perhaps @jcrist can weigh in?

aktech:
@fanshi118 @jacobtomlinson Is there much left to do in here? Is there anything I can help with?

jacobtomlinson:
@aktech I think things are pretty good here. Looks like we were just blocked on the decision to push artifacts for every commit. My vote on this would be no.

martindurant:
This is still picking up travis, so we definitely at least need to merge from recent

consideRatio:
I'll go for a merge on this and use it as a starting point to iterate from!

PR #353, opened 2020-11-26
Add customizable worker_threads

### LINKED ISSUE #362: Allow single-threaded running of workers
# Missing feature
I don't find any way to run the Dask workers single-threaded using `dask-gateway` (on Kubernetes). This is required because the Python package used on the workers by me is not thread-safe (some old underlying FORTRAN-code). This is certainly a common use case.

# Context
`dask` / `distributed` themselves offers free configuration to allow multi-threading or not: https://docs.dask.org/en/latest/scheduling.html

So it is just something like this:

```
from dask.distributed import Client

client = Client(address=SCHEDULER_URI, processes=True, n_workers=NUM_LOCAL_WORKERS, threads_per_worker=1)
```

However, if using `dask-gateway`, I do not find any hint on how to do this. In fact, if searching the documentation, there is no single hit for "thread" or similar keywords. A question on Stackoverflow did not get any results: https://stackoverflow.com/questions/65107333/how-to-run-single-threaded-workers-with-dask-gateway

Actually, the `dask-gateway` is starting the workers multi-threaded:
```
from dask_gateway import GatewayCluster

cluster = GatewayCluster(address=cluster_gateway_address)
client = cluster.get_client()
```

This is resulting in:
```
distributed.nanny - INFO -         Start Nanny at: 'tls://10.0.0.84:42257'
distributed.worker - INFO -       Start worker at:      tls://10.0.0.84:36645
distributed.worker - INFO -          Listening to:      tls://10.0.0.84:36645
distributed.worker - INFO -          dashboard at:             10.0.0.84:8787
distributed.worker - INFO - Waiting to connect to: tls://dask-64c3b5e1bdca47749bcce9572481daa9.default:8786
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -               Threads:                          2
distributed.worker - INFO -                Memory:                    3.22 GB
distributed.worker - INFO -       Local Directory: /home/dask/dask-worker-space/dask-worker-space/worker-vd_dpnn0
distributed.worker - INFO - -------------------------------------------------
```

I tried this, but it does not seem to work:
```
from dask_gateway import GatewayCluster

cluster = GatewayCluster(address=CLUSTER_GATEWAY_URL, processes=True, n_workers=NUM_WORKERS, threads_per_worker=1)
client = cluster.get_client()
```

It results in:
```
ValueError: Unknown fields ['n_workers', 'processes', 'threads_per_worker']
```

# Suggested solution
It should be possible to configure `dask-gateway` regarding threads/processes in the same way as `dask` and `distributed`, at least multi-threading should be possible to switch off.

Or is this already possible right now? In this case the documentation should clearly describe this.

/Ralf

### LINKED ISSUE #364: [kubernetes] Cannot specify threads per worker
[This line](https://github.com/dask/dask-gateway/blob/e409f0e87f45e0a51fd7c009b5ec010bc5253bf1/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py#L1055) fixes the threads per worker to the cpu request. There are use cases where you might not want this behavior. For example, say you are mapping a workflow that requires a lot of memory and can make use of other multithreading functionality (e.g. OpenMP) across multiple inputs. You may want your pod to have, say, 4 CPUs. But you may want to allow OpenMP to use 4 threads while only letting dask control a single thread. As the code stands right now, you have to enable 4 dask threads and thus your big workflow would run out of memory trying to run on 4 inputs simultaneously.

Would it make sense to add in a ``threads_per_worker`` config option, which defaults to being equal to `str(int(config.worker_cores_limit))`?

I'm mentioning this in the Kubernetes context b/c that's the backend we're using when running into this issue, but you might want it for other backends as well?

AndreaGiardini:
Hey

I added the possibility to customize the number of threads per worker.
Until now `worker_threads` = `cores` but that's not always the best combination.

If `worker_threads` is not set then it will fall-back to the default.

Let me know if you have any feedback

Fixes #364 
Fixes #362

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -1052,7 +1052,7 @@ class KubeController(KubeBackendAndControllerMixin, Application):
             "--name",
             "$(DASK_GATEWAY_WORKER_NAME)",
             "--nthreads",
-            str(int(config.worker_threads)),
+            str(config.worker_threads),
             "--memory-limit",
             str(config.worker_memory_limit),
         ]
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -1052,7 +1052,7 @@ class KubeController(KubeBackendAndControllerMixin, Application):
             "--name",
             "$(DASK_GATEWAY_WORKER_NAME)",
             "--nthreads",
-            str(int(config.worker_cores_limit)),
+            str(int(config.worker_threads)),
             "--memory-limit",
             str(config.worker_memory_limit),
         ]
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ jobs:
             - source continuous_integration/before_install.sh
           install:
             - source continuous_integration/install.sh
-            - pip install kubernetes_asyncio skein sphinx doctr dask-sphinx-theme sphinx-rtd-theme==0.4.3
+            - pip install kubernetes_asyncio skein sphinx==3.3.1 doctr dask-sphinx-theme sphinx-rtd-theme==0.4.3
           script:
             - |
               set -xe
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -279,6 +279,16 @@ class ClusterConfig(Configurable):
         config=True,
     )
 
+    # Number of threads per worker. Defaults to the number of cores
+    worker_threads = Integer(
+        1,
+        min=1,
+        help="""
+        Number of threads available for a dask worker.
+        """,
+        config=True,
+    )
+
     scheduler_memory = MemoryLimit(
         "2 G",
         help="""
--- a/dask-gateway-server/dask_gateway_server/backends/base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/base.py
@@ -281,13 +281,18 @@ class ClusterConfig(Configurable):
 
     # Number of threads per worker. Defaults to the number of cores
     worker_threads = Integer(
-        1,
         min=1,
         help="""
         Number of threads available for a dask worker.
+
+        Defaults to ``worker_cores``.
         """,
         config=True,
     )
+
+    @default("worker_threads")
+    def _default_worker_threads(self):
+        return self.worker_cores
 
     scheduler_memory = MemoryLimit(
         "2 G",
--- a/dask-gateway-server/dask_gateway_server/backends/db_base.py
+++ b/dask-gateway-server/dask_gateway_server/backends/db_base.py
@@ -1443,7 +1443,7 @@ class DBBackendBase(Backend):
         ]
 
     def worker_nthreads_memory_limit_args(self, cluster):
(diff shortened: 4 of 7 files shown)


lorenzori (approved):

AndreaGiardini:
Hey @jcrist 
I had a look at the failing CI and it looks like `dask/distributed` does not communicate the number of cores but only the nthreads, this is why the tests for the widget fail:

https://github.com/dask/distributed/blob/b4dfc925bac32a488be2016a5930a9b7dd95cec5/distributed/scheduler.py#L1707-L1728

Do you have any preference on how to tackle this?

AndreaGiardini:
Friendly ping to @consideRatio . Maybe you can give me some feedback on this?

erl987:
Very useful initiative relates to issue #362.

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -1052,7 +1052,7 @@ def get_worker_command(self, namespace, cluster_name, config):
             "--name",
             "$(DASK_GATEWAY_WORKER_NAME)",
             "--nthreads",
-            str(int(config.worker_cores_limit)),
+            str(int(config.worker_threads)),
```
consideRatio, line 1055: I think worker_threads is of an `Integer` type already ([traitlets](https://traitlets.readthedocs.io/en/stable/)), while worker_cores_limit was a `Float` traitlet type and needed the cast to integer first.

```suggestion
            str(config.worker_threads),
```

consideRatio (commented):
Thanks for your work @AndreaGiardini! I'm not representing this project but hopefully my review can be helpful.

First of all I think its sensible to allow worker_threads to be customizable, so the feature idea is sound.

What I would suggest iterating on is mainly the default value of worker_threads in the normal ClusterConfig and in the KubeClusterConfig.

The ClusterConfig has a default value of worker_threads to be 1, but in the past, it was matching the amount of CPU. The KubeClusterConfig deriving from ClusterConfig has a dynamic default value of worker_threads set to be worker_cores, but in the past int(worker_cores_limit) was used.

As the PR summary or title doesn't indicate an intent to change the default behavior, I think what makes sense to recommend, is to update the PRs code suggestion to implement a dynamic _default_worker_threads function with the `@default` decorator in both ClusterConfig and KubeClusterConfig with the different defaults previously used.

AndreaGiardini:
@consideRatio Thank you for the valuable input

I fixed the PR as you suggest, I can always squash as final step before merging if needed.
Still, the problem remains with the widget unfortunately... How do you think we should tackle this?

AndreaGiardini:
Hey @consideRatio and @jcrist 

I took some time to rebase the PR and make CI green without upgrading the distributed dep.
I found an issue with the docs generation, that should be fixed as well.

Since there seems to be quite some interest in this feature, can you tell me if something else is needed to get this merged? Thank you

martindurant:
@consideRatio , are you happy with the changes here?

This all looks fair as far as I can tell.

@jcrist , are you likely to have time to have a quick look?

@jacobtomlinson , this is one of the few repos still on Travis, which still appears to work for now. I expect it might be more complex than most to transform.

jcrist:
Apologies for letting this linger, this looks good to me. Merging. Thanks for the contribution!

martindurant:
Thanks, @jcrist

AndreaGiardini:
Thank you to all of you! :) Would you mind tagging a new release whenever possible? We could really benefit from this

consideRatio:
Wieee! :tada:

PR #365, opened 2020-12-25
The function of accessing clusters in the form of load balance

HamzaPekdogan:
We encountered a situation where multi cluster should work. As a solution, we felt the need to add a new function. We wanted to meet the load balance need.


consideRatio:
@HamzaPekdogan oh hmmm, I don't understand this well enough.

If you are still interested in proposing a feature related to this, please open an issue where you describe the use case and help me or another maintainer understand it better.

HamzaPekdogan:
@consideRatio When you do get_cluster, it is always run on the same dask cluster. I made it work on a different dask cluster each time by doing round robin here.

HamzaPekdogan:
@jrbourbeau Can you review this work of mine?

Issue #389, opened 2021-04-15
Templatize the kubernetes resources that dask-gateway generates


costrouc:
For QHub we have moved away from using the dask-gateway helm chart to more tightly integrate dask-gateway with our traefik http/https/tcp proxy. This was mainly motivated when we found of that two traefik services in the same namespace in kubernetes do not play well with each other.

All that said the current issues we are facing is around decorating the `IngressRoute` and needing to add `tls: {"certManager": "default"}`. I would like to propose templatizing the [resource objects being created](https://github.com/dask/dask-gateway/blob/7d1659db8b2122d1a861ea820a459fd045fc3f02/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py#L1192-L1295) via Traitlets. 

For example

```python
INGRESS_ROUTE_TEMPLATE = {
            "apiVersion": "traefik.containo.us/v1alpha1",
            "kind": "IngressRoute",
            "metadata": {
                "labels": "PLACEHOLDER",
                "annotations": "PLACEHOLDER",
                "name": "PLACEHOLDER",
            },
            "spec": {
                "entryPoints": "PLACEHOLDER",
                "routes": [
                    {
                        "kind": "Rule",
                        "match": "PLACEHOLDER",
                        "services": [
                            {
                                "name": "PLACEHOLDER",
                                "namespace": namespace,
                                "port": 8787,
                            }
                        ],
                        "middlewares": "PLACEHOLDER",
                    }
                ],
            },
        }
```

Or possibly we should just make the `make_ingressroute` functions and similar overridable via traetlets callables. We need this functionality to expose the dask scheduler dashboard with `https`. 

cc: @aktech

TomAugspurger:
Thanks! I think it'd be great to remove the need for this workaround in QHub.

@droctothorpe or @consideRatio does this proposal sound sensible to you?

consideRatio:
### My understanding summarized

- dask-gateway the Helm chart creates traefik.containo.us/v1alpha1/IngressRoute resources
- These custom resources can be managed by Traefik, and dask-gateway deploys Traefik
- Some IngressRoute resources are created directly by the dask-gateway Helm chart templates, and some are dynamically created by the DaskCluster controller for each DaskCluster resource
- Those IngressRoute resources that are dynamically created, are created by the function `make_ingressroute`.

I'll number some of my thoughts as I consider this further.

### Questions in my mind

#### Better understanding of the problem
[...]

PR #392, opened 2021-04-29
Make the traefik ingressroute and ingressroutetcp resources configurable

costrouc:
With QHub we have the goal of deploying kubernetes on multiple clouds and locally which makes depending on the load balancer for encryption difficult. We needed to annotate (somehow) the ingressroute resource to add tls via the following snippet. More details are in the [issue](https://github.com/dask/dask-gateway/issues/389). Here I've tried to make this as uncontroversial as possible by just adding two Callables for the `make_ingressroute` and `make_ingressroutetcp`.

```yaml
tls:
  certResolver: default
```

Currently with dask-gateway it is not possible to override the traefik ingressroute and ingresstls resource being generated on the fly by the KubeController.

Closes https://github.com/dask/dask-gateway/issues/389

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -347,13 +347,13 @@ class KubeController(KubeBackendAndControllerMixin, Application):
     make_ingressroute = Callable(
         _make_ingressroute,
         help="A callable function to override the default traefik ingressroute resource",
-        config=True
+        config=True,
     )
 
     make_ingressroutetcp = Callable(
         _make_ingressroutetcp,
         help="A callable function to override the default traefik ingressroutetcp resource",
-        config=True
+        config=True,
     )
 
     _log_formatter_cls = LogFormatter
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -16,6 +16,7 @@ from kubernetes_asyncio import client, config
 from kubernetes_asyncio.client.rest import ApiException
 
 from ... import __version__ as VERSION
+from ...traitlets import Callable
 from ...utils import (
     AccessLogger,
     LogFormatter,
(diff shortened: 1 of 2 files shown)


costrouc:
If it helps the errors here are the same on the `main` branch https://travis-ci.org/github/dask/dask-gateway/jobs/762173118

costrouc:
This PR is tested and currently being used with QHub so I can give assurance that this PR works.

costrouc:
Rebased on main to hopefully make CI pass 🙏

martindurant:
> hopefully make CI pass

Not quite :|

costrouc:
I think the tests for pbs are flaky since this does not touch those tests

martindurant:
OK, t does pass. I'm not sure who can best review this - looks OK to me, but I don't really TLS. I don't think you're actually testing the non-default option, which essentially does the same thing as before.

@aktech , I don't know if you have any plans to look into flaky tests.

aktech:
> @aktech , I don't know if you have any plans to look into flaky tests.

Not immediately, but sometime soon for sure.

martindurant:
@costrouc , my only request here was that it's be great if you could test changing the new configuration to be non-default.

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -285,6 +344,18 @@ def _validate_proxy_prefix(self, proposal):
...
+        help="A callable function to override the default traefik ingressroute resource",
+        config=True,
+    )
+
+    make_ingressroutetcp = Callable(
+        _make_ingressroutetcp,
+        help="A callable function to override the default traefik ingressroutetcp resource",
+        config=True,
+    )
+
```
consideRatio, line 358: I'd like a brief iteration on this helm string. This configuration is to override the default instance function on the KubeController class that takes a Dask cluster name and a namespace and returns a string representing a k8s manifest.

I guess what I'm missing most from the current description is the fact that it should be a callable function taking `self` (KubeController), clustername, namespace, and returns a string representing a valid k8s resource manifest.

costrouc:
Closing since we found that the proper way to accomplish this is to set tls settings on the entrypoints.

PR #396, opened 2021-05-24
added a missing comma in the profile options code chunk example

cdibble:
This fixes an example code chunk for exposing cluster options as profiles. The missing comma caused a syntax error on deployment.


jcrist:
Thanks!

Issue #398, opened 2021-06-01
dask-gateway-server proxy fails to build with Go 1.16


douglasdavis:
**What happened**:

```
pip install -e ./dask-gateway-server
```
fails with Go 1.16 installed.

<details>
<summary>Full failure</summary>
<br>

```
  Running setup.py develop for dask-gateway-server
    ERROR: Command errored out with exit status 1:
     command: /home/ddavis/.pyenv/versions/3.9.5/envs/dask-dev/bin/python3.9 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/ddavis/software/repos/dask-gateway/dask-gateway-server/setup.py'"'"'; __file__='"'"'/home/ddavis/software/repos/dask-gateway/dask-gateway-server/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
         cwd: /home/ddavis/software/repos/dask-gateway/dask-gateway-server/
    Complete output (5 lines):
    running develop
    running build_proxy
    go: cannot find main module, but found .git/config in /home/ddavis/software/repos/dask-gateway
    	to create a module there, run:
    	cd ../.. && go mod init
    ----------------------------------------
ERROR: Command errored out with exit status 1: /home/ddavis/.pyenv/versions/3.9.5/envs/dask-dev/bin/python3.9 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/home/ddavis/software/repos/dask-gateway/dask-gateway-server/setup.py'"'"'; __file__='"'"'/home/ddavis/software/repos/dask-gateway/dask-gateway-server/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps Check the logs for full command output.
```
</details>

**What you expected to happen**:

A clean installation

**Minimal Complete Verifiable Example**:

Install Go 1.16+

```
[...]

PR #399, opened 2021-06-04
Add warning about Go version to server installation instructions.

douglasdavis:


douglasdavis:
This is a band-aid on https://github.com/dask/dask-gateway/issues/398

martindurant:
(keeping the issue open, since the env variable should not be required)

PR #415, opened 2021-07-29
Add GitHub Actions badges in README.md

aktech:
replacing travis with GitHub Actions

cc @martindurant

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 dask-gateway
 ============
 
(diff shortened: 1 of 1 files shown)


martindurant:
Do you have a sense for how flaky the test runs are likely to be?

aktech:
It's mainly pbs and slurm tests. The former fails like 1 in 8 and the latter 1 in maybe 12 of what I have seen. Everything else seems fine. I haven't had the chance to take a deeper look on the root cause though.

martindurant:
When things run fine locally but fail on CI, it's usually timeouts - everything runs much slower, and there can be unexpected times when nothing is happening at all.

PR #416, opened 2021-08-05
Use format_bytes from dask instead of distributed

TomAugspurger:
https://github.com/dask/distributed/pull/4966 deprecated
`distributed.utils.format_bytes`. We should use the version from
`dask.utils`. I didn't check the exact version of Dask it was added, but
it was >2 years ago, so I think we're OK.

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -15,10 +15,11 @@ from urllib.parse import urlparse
 import aiohttp
 import dask
 import yarl
+from dask.utils import format_bytes
 from distributed import Client
 from distributed.core import rpc
 from distributed.security import Security
-from distributed.utils import LoopRunner, format_bytes
+from distributed.utils import LoopRunner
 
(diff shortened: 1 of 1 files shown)


jrbourbeau (commented):
Thanks @TomAugspurger! This look good. I noticed `dask_gateway_server` has its own `format_bytes` utility:

https://github.com/dask/dask-gateway/blob/7d1659db8b2122d1a861ea820a459fd045fc3f02/dask-gateway-server/dask_gateway_server/utils.py#L159

that can probably also be removed in favor of using Dask's directly

TomAugspurger:
I might be wrong, but I *think* that dask-gateway-server doesn't depend on dask / distributed at all (https://github.com/dask/dask-gateway/blob/7d1659db8b2122d1a861ea820a459fd045fc3f02/dask-gateway-server/setup.py#L100).

jrbourbeau:
Whoops, in that case nvm : )

PR #420, opened 2021-08-21
Helm chart: update deprecated k8s resources no longer supported in k8s 1.22

### LINKED ISSUE #419: Helm chart: transition apiextensions.k8s.io/v1beta1 to v1 (required for use with k8s 1.22)
As described in [the v1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122) deprecation guide, the CustomResourceDefinition resources in this Helm chart relies on the `apiVersion: apiextensions.k8s.io/v1beta1` that has been removed in k8s 1.22. This makes this Helm chart no longer install properly in k8s 1.22.

I'm already started working to fix this.

### Related
- We seem to have imported the CRDs from Traefik, [as defined here](https://github.com/traefik/traefik-helm-chart/tree/master/traefik/crds), and they have been updated already upstream [in this commit](https://github.com/traefik/traefik-helm-chart/commit/3a3396892aec5c742fafeaeb15b0d9ecc81bb9e4).

consideRatio:
### PR summary
- PR closes #419 by updating the deprecated k8s API `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`
  - Re-imports [generated files from traefik/traefik-helm-chart](https://github.com/traefik/traefik-helm-chart/tree/master/traefik/crds)
  - Manually updates a k8s manifest for the DaskCluster resource definition
- The helm chart no longer supports k8s <=1.15
  - We can remain backward compatible with extra work, but I don't consider it worth doing. As comparison, the JupyterHub Helm chart requires k8s 1.17+ atm.
- The helm chart now supports k8s >=1.22
- A very efficient test running `helm lint` and `helm template` is added before the heavyweight test to build images and install the Helm chart is ran
  image

### Review notes

I have been a bit cautious to put in work in this repo because I'm not sure if a PR I'd make would be reviewed -> merged given limited maintenance capacity available among dask maintainers. If it would help, I'd be very happy to take on some maintenance responsibility and issue triage/review/merge PRs related to the dask-gateway Helm chart and/or documentation in general.

I'm not sure who has rights to make decisions about these matters or can review this PR, perhaps @martindurant or @jacobtomlinson?

Thanks everyone involved for an amazing project!

### DIFF
--- a/continuous_integration/kubernetes/helm-lint-and-template.sh
+++ b/continuous_integration/kubernetes/helm-lint-and-template.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+set -e
+
+this_dir="$(dirname "${BASH_SOURCE[0]}")"
+git_root="$(cd "${this_dir}/../.." && pwd)"
+
+echo "Linting Helm Chart"
+helm lint \
+    "${git_root}/resources/helm/dask-gateway" \
+    -f "${git_root}/resources/helm/testing/travis.yaml"
+
+echo "Rendering Helm Chart Templates"
+helm template dask-gateway \
+    --include-crds \
+    "${git_root}/resources/helm/dask-gateway" \
+    -f "${git_root}/resources/helm/testing/travis.yaml"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -87,6 +87,10 @@ jobs:
           k3d --version
           kubectl get nodes
 
+      - name: Helm lint and render templates
+        run: |
+          ./continuous_integration/kubernetes/helm-lint-and-template.sh
+
       - name: Helm Install
         run: |
           stern "" > k8s-logs &
--- a/continuous_integration/kubernetes/install-tools.sh
+++ b/continuous_integration/kubernetes/install-tools.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
-KUBE_VERSION=1.16.0
-HELM_VERSION=3.1.2
+KUBE_VERSION=1.22.1
+HELM_VERSION=3.6.3
 STERN_VERSION=1.11.0
 
 # Install kubectl
--- a/resources/helm/dask-gateway/Chart.yaml
+++ b/resources/helm/dask-gateway/Chart.yaml
@@ -7,4 +7,5 @@ description: A multi-tenant server for deploying and managing Dask clusters
 home: https://gateway.dask.org/
 sources:
   - https://github.com/dask/dask-gateway/
+icon: https://docs.dask.org/en/latest/_images/dask_icon_no_pad.svg
 kubeVersion: ">=1.16.0-0"
--- a/resources/helm/dask-gateway/Chart.yaml
+++ b/resources/helm/dask-gateway/Chart.yaml
@@ -1,3 +1,4 @@
+# Chart.yaml v2 reference: https://helm.sh/docs/topics/charts/#the-chartyaml-file
 apiVersion: v2
 name: dask-gateway
 version: 0.9.0
@@ -6,3 +7,4 @@ description: A multi-tenant server for deploying and managing Dask clusters
 home: https://gateway.dask.org/
 sources:
   - https://github.com/dask/dask-gateway/
+kubeVersion: ">=1.16.0-0"
--- a/resources/helm/dask-gateway/crds/daskclusters.yaml
+++ b/resources/helm/dask-gateway/crds/daskclusters.yaml
@@ -1,15 +1,31 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: daskclusters.gateway.dask.org
+  creationTimestamp: null
 spec:
   group: gateway.dask.org
-  version: v1alpha1
   names:
     kind: DaskCluster
     listKind: DaskClusterList
     plural: daskclusters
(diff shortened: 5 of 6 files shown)


jacobtomlinson:
Thanks for your efforts here @consideRatio! I'll take a look.

> I'd be very happy to take on some maintenance responsibility and issue triage/review/merge PRs related to the dask-gateway Helm chart and/or documentation in general.

I'm sure this would be much appreciated! I think we would need to get a approval from a few Dask maintainers to make this happen. I'll raise it at the weekly meeting tomorrow.

continuous_integration/kubernetes/helm-lint-and-template.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -e
+
+this_dir="$(dirname "${BASH_SOURCE[0]}")"
+git_root="$(cd "${this_dir}/../.." && pwd)"
+
+echo "Linting Helm Chart"
+helm lint \
```
jacobtomlinson, line 8: I wonder if we could use precommit here instead of a CI script to keep things consistent with `dask/helm-chart`.

https://github.com/dask/helm-chart/blob/main/.pre-commit-config.yaml#L6-L9

jacobtomlinson (commented):
This generally looks great thanks!

continuous_integration/kubernetes/helm-lint-and-template.sh
consideRatio: From inspection, the referenced GitHub action to do `helm lint` [don't support referencing our test config in the current location](https://github.com/gruntwork-io/pre-commit/blob/e9250bd69bb312d55364213ff5ff037a09be55d9/hooks/helmlint.sh#L113-L120).

I suggest if there is a pre-commit hook to be added about helm lint, that it is done in a separate PR in addition to this which references a specific config file referenced by other tests.

consideRatio:
Thanks you for your review work @jacobtomlinson! :tada: :heart:

continuous_integration/kubernetes/helm-lint-and-template.sh
jacobtomlinson: > I suggest if there is a pre-commit hook to be added about helm lint, that it is done in a separate PR in addition to this which references a specific config file referenced by other tests.

I understand where you are coming from, but this PR adds a listing script, and it would be preferable to do this via precommit to be consistent with other places.

If we bump precommit to a separate PR then that PR would also remove this script right?

consideRatio: This script includes `helm lint` and `helm template`. The key thing I wanted to add was actually `helm template` which is what I find to typically catch most template issues. So by using the dedicated `helm lint` pre-commit action by gruntworks-io and making pre-commit run as part of the CI system, it would not solve the wish to have a `helm template` step here.

Here are options I could execute on directly:
1. Remove `helm lint` from this script:
   a) and add a pre-commit hook to do `helm lint`
   b) and don't do `helm lint` in the CI system
2. Remove this script entirely and do a local pre-commit hook to do both `helm template` as well as `helm lint`
3. Stick to not referencing `helm` in pre-commit but instead doing it only in the GitHub Workflows

I generally dislike the complexity added by using a third party action to do something as minor as running `helm lint` because of how it makes it harder to understand whats going on, so in my order of preference if we are to use pre-commit would be: 2 -> 1b -> 1a, but due to an experience described below, I suggest 3.

### Using pre-commit.ci

In the JupyterHub org, we have went all out on pre-commit and started using also the pre-commit.ci service. It helps to:
- update versions referenced in pre-commit-config.yaml files
- push pre-commit changes to PRs that fail tests
- run pre-commit checks in a very quick and performant manner and report its status like any other GitHub workflow job

If pre-commit need to reference the `helm` binary directly, as `helm template` would, it would mean that there is a dependency on that binary for CI systems to run. Local users would need to install `helm` even though they don't work with the Helm chart but instead work on dask-gateway's Python code base, and pre-commit.ci wouldn't be able to handle this either.

For reasons like this, we have [explicitly removed `helm` in the JupyterHub helm chart's pre-commit config](https://github.com/jupyterhub/zero-to-jupyterhub-k8s/commit/c37d5f79cb838df5c8e94677f8f6b5b0a27fb804#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9). Even if we would reference gruntworks-io, it [would require `helm` to be available](https://github.com/gruntwork-io/pre-commit/blob/e9250bd69bb312d55364213ff5ff037a09be55d9/hooks/helmlint.sh#L122-L132) and fail without it being installed.

---

@jacobtomlinson let me know what you would like me to do and I'll get it done!

TomAugspurger (approved):
Thanks for fixing this Erik. I'll let Jacob comment on the CI approach. I'm fine with either (and the experience from jupyterhub is very helpful.).

continuous_integration/kubernetes/helm-lint-and-template.sh
jacobtomlinson: Thanks for taking the time to write this up @consideRatio. That's compelling enough for me. Let's go with this as is.

consideRatio: Thanks for the review and patience with me on this @jacobtomlinson!

jcrist:
Thanks for the excellent PRs @consideRatio! You should now have write access to this repository.

consideRatio:
Thank you @jacobtomlinson, @TomAugspurger, @jcrist! :tada:

Issue #421, opened 2021-08-21
Helm chart: make images arm64 compatible and publish images supporting both amd64 and arm64


consideRatio:
A container registry can have multiple images under the same name for different CPU architectures. Thanks to that, the same Helm chart for example, can reference the same image name, and be installable both on amd64 machines and arm64 based machines.

I'd like for this Helm chart to be installable in arm64 based environments as well, for example on a RaspberryPi based k8s cluster.

Practically, this can be done by using `buildx` which is like `docker build` but you can then also pass `--platforms linux/amd64,linux/arm64` and voila your images will be both amd64 and arm64 compatible. This has been done for JupyterHub for example and is very common among Helm charts. Amd64 and Arm64 are the most popular CPU architectures to support.

A reference implementation setup can be found here: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/d848959b6229d3adaf2cea7e06f79cac1433fc43/.github/workflows/test-docker-build.yaml#L41-L54

### Related
- #423 makes the Dockerfiles successfully build on arm64 platforms

consideRatio:
I've worked to make this possible, and we are almost there I think. Right now, the only issue is that psutils is failing to build in arm64 without installing a lot of extra stuff making things complicated. I'd rather help psutils support building an arm64 wheel than workaround it though.

consideRatio:
I'll close this as resolved for now, and track the need to handle this in #543.

PR #423, opened 2021-08-21
images: some refactoring and version bumps for arm64 compatible Dockerfiles

consideRatio:
This is a step along the way for #421 about compatibility on arm64 machines like Raspberry Pi.

### DIFF
--- a/dask-gateway-server/Dockerfile
+++ b/dask-gateway-server/Dockerfile
@@ -1,18 +1,19 @@
-FROM python:3.8.6-slim-buster as dependencies
+FROM python:3.8.11-slim-buster as dependencies
 LABEL MAINTAINER="Jim Crist-Harif"
 
 RUN apt-get update \
     && apt-get install -y tini \
-    && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
 RUN pip install --no-cache-dir \
-    aiohttp==3.7.2 \
+    aiohttp==3.7.4 \
     colorlog \
     cryptography \
-    traitlets==4.3.3 \
+    traitlets==5.0.5 \
     pyyaml \
-    kubernetes-asyncio==11.0.0
+    kubernetes-asyncio==12.1.1
+
+
 
 # Build dask-gateway-server from source in a builder stage
 FROM dependencies AS builder
@@ -26,6 +27,8 @@ RUN python setup.py install \
     --single-version-externally-managed \
     --record=record.txt \
     --prefix /tmp/install-prefix
+
+
 
 # Final image - merge dependencies and built dask-gateway
 FROM dependencies
--- a/dask-gateway/Dockerfile
+++ b/dask-gateway/Dockerfile
@@ -2,27 +2,34 @@
 FROM debian:buster-slim as miniconda
 LABEL MAINTAINER="Jim Crist-Harif"
 
-ARG CONDA_VERSION=py38_4.8.3
-ARG CONDA_SHA256=879457af6a0bf5b34b48c12de31d4df0ee2f06a8e68768e5758c3293b2daf688
+# List of conda versions: https://repo.anaconda.com/miniconda/
+#
+# Miniconda 4.9.2 has been chosen as it has a aarch64 release and hasn't
+# introduced a bug described in:
+# https://stackoverflow.com/questions/68213186/illegal-instruction-error-when-verifying-anaconda-miniconda-install
+#
+# FIXME: Use micromamba or mambaforge instead, see
+#        https://github.com/mamba-org/mamba for more info
+#
+ARG CONDA_VERSION=py38_4.9.2
 
 # - Create user dask
+RUN useradd -m -U -u 1000 dask
+
 # - Install tini
 # - Install miniconda build dependencies
-# - Download miniconda and check the sha256 checksum
-# - Install miniconda
+RUN apt-get update \
+    && apt-get install -y tini wget bzip2 \
+    && rm -rf /var/lib/apt/lists/*
+
+# - Download and install miniconda
 # - Configure conda to minimize automatic package updates
-# - Remove all conda managed static libraries
 # - Cleanup conda files
 # - Uninstall miniconda build dependencies
-RUN useradd -m -U -u 1000 dask \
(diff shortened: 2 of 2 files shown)


dask-gateway-server/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
 RUN pip install --no-cache-dir \
-    aiohttp==3.7.2 \
+    aiohttp==3.7.4 \
     colorlog \
     cryptography \
-    traitlets==4.3.3 \
+    traitlets==5.0.5 \
     pyyaml \
-    kubernetes-asyncio==11.0.0
+    kubernetes-asyncio==12.1.1
```
consideRatio, line 14: Systematic version updates as I had to update aiohttp in order to have a arm64 (aka. aarch64) compatible conda package in the dask-gateway image.

dask-gateway/Dockerfile
```diff
@@ -2,27 +2,34 @@
 FROM debian:buster-slim as miniconda
 LABEL MAINTAINER="Jim Crist-Harif"
 
-ARG CONDA_VERSION=py38_4.8.3
-ARG CONDA_SHA256=879457af6a0bf5b34b48c12de31d4df0ee2f06a8e68768e5758c3293b2daf688
```
consideRatio, line 6: Removal of checksums as amd64 / arm64 will have different checksums and it is a bit tedious to keep track of both.

dask-gateway/Dockerfile
```diff
@@ -2,27 +2,34 @@
...
 # - Create user dask
+RUN useradd -m -U -u 1000 dask
+
 # - Install tini
 # - Install miniconda build dependencies
-# - Download miniconda and check the sha256 checksum
-# - Install miniconda
+RUN apt-get update \
+    && apt-get install -y tini wget bzip2 \
+    && rm -rf /var/lib/apt/lists/*
```
consideRatio, line 23: I split this apart from a section below making so many different things in the same RUN layer that it became a bit hard to read.

dask-gateway/Dockerfile
```diff
@@ -32,39 +39,40 @@ RUN useradd -m -U -u 1000 dask \
     && echo "aggressive_update_packages: []" >> /home/dask/.condarc \
     && find /opt/conda/ -follow -type f -name '*.a' -delete \
     && /opt/conda/bin/conda clean -afy \
-    && chown -R dask:dask /opt/conda \
-    && apt-get autoremove --purge -y wget bzip2 \
```
consideRatio, line 36: By not removing wget / bzip2 the image will not grow further because adding + removing won't help if its not done in the same layer, and since it is no longer in the same layer I don't remove these binaries later.

The reason for having one more RUN statements was for readability.

dask-gateway/Dockerfile
```diff
@@ -32,39 +39,40 @@ RUN useradd -m -U -u 1000 dask \
...
 
+
+
 # ** An image with all of dask-gateway's dependencies **
 FROM miniconda as dependencies
 
-ARG DASK_VERSION=2021.7.1
-ARG DISTRIBUTED_VERSION=2021.7.1
+ARG DASK_VERSION=2021.8.1
+ARG DISTRIBUTED_VERSION=2021.8.1
```
consideRatio, line 56: I just bumped these while I was at it so all the conda/pip packages installed were updated across now that I had to update at least aiohttp.

dask-gateway/Dockerfile
```diff
@@ -32,39 +39,40 @@ RUN useradd -m -U -u 1000 dask \
...
-        pandas==1.1.4 \
+        numpy==1.21.2 \
+        pandas==1.3.2 \
     && /opt/conda/bin/conda clean -afy \
     && find /opt/conda/ -follow -type f -name '*.a' -delete \
     && find /opt/conda/ -follow -type f -name '*.js.map' -delete \
     && find /opt/conda/lib/python*/site-packages/bokeh/server/static -follow -type f -name '*.js' ! -name '*.min.js' -delete
 
+
+
```
consideRatio, line 75: For readability as separation between the build steps

jacobtomlinson (commented):
This generally looks good to me!

We recently added Arm support to the Dask docker images in dask/dask-docker#166.

I wonder if there are any learnings we can take from that PR, or anything we can reuse to ensure a bit more consistency between projects.

consideRatio:
> We recently added Arm support to the Dask docker images in [dask/dask-docker#166](https://github.com/dask/dask-docker/pull/166).
> 
> I wonder if there are any learnings we can take from that PR, or anything we can reuse to ensure a bit more consistency between projects.

Ah one key thing to keep consistent is a transition to mamba and mambaforge to install it. I'd be happy to help with that as well but it felt like out of scope for this PR to do that also.

dask-gateway/Dockerfile
TomAugspurger: Do you have a rough idea of how much larger the image is now?

TomAugspurger (approved):
Thanks for this, one question about image sizes inline.

I'd love it if we could come up with a policy about what library versions to use in the Dockerfiles, to avoid having repeated conversations about versions. Is there any reason *not* to use the latest version of all libraries (dask, distributed, aiohttp, etc) at the time of release?

One maybe special consideration is Python itself. Right now we're on 3.8.6 (bumping to 3.8.11). Is there any reason to keep that on 3.8? Or should we move to 3.9?

consideRatio:
I think a good policy is to use a requirements.in file that is automatically refrozen to requirements.txt by dependabot, and then we install such requirements.txt in the dockerfile. That helps users get info about what version has been used in each release via git history, we avoid a maintenance burden, and we get security warnings via automation etc.

We have a mix of conda/pip in these files though, making my idea a bit troublesome. I'd consider going 100% pip in these images or 100% conda and hope there is a good tool like there is for pip that dependabot can use for an environment.yml file

consideRatio:
Go for merge on this?

jcrist (approved):
Thanks @consideRatio!

consideRatio:
Thanks @jacobtomlinson @TomAugspurger @jcrist for your review efforts!

PR #425, opened 2021-08-21
helm chart: refactor to use consistent modern syntax

consideRatio:
I recognize a lot of practices from the JupyterHub Helm chart made it into this Helm chart. In this PR make the templates consistent in its syntax across helm template files and align with the best practices that have now also to some degree made its way all to Helms official documentation.

This PR should have no influence on anything in practice.

About the various kinds of changes made:
- `trimSuffix "\n"` is no longer needed with modern versions of `helm` and would be harmless to remove anyhow
- A replica count had a `default 1` that served no purpose
- Sometimes it was `. | toYaml | nindent 4`, othertimes it was `toYaml . | nindent 4`. I made it consistently like the former example.
- Some whitespace chomping practices enforced:
  - always chomp right on the first line in a template file
  - almost always chomp left but not to the right elsewhere
- if statements became with statements where possible to avoid duplicating a reference to some variable (with statements also function as as if statements).
- redundant parenthesis removed

### DIFF
--- a/resources/helm/dask-gateway/templates/controller/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/controller/deployment.yaml
@@ -20,8 +20,8 @@ spec:
         app.kubernetes.io/component: controller
       annotations:
         checksum/configmap: {{ include (print .Template.BasePath "/controller/configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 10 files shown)


jacobtomlinson (approved):
Awesome thanks so much!

PR #431, opened 2021-08-26
Helm chart: update to traefik v2.5.x

consideRatio:
I bumped Traefik to the latest version and unpinned the patch version of the image to `v2.5` which I find reasonable at least until we have more regular releases.

To make traefik start and function properly, I had to also install a few CRDs and grant Traefik permission to inspect those CRDs. Specifically, I've added the following CRDs and associciated read permissions (get, list, watch).
- IngressRouteUDP
- MiddlewaresTCP
- ServersTransports
- TLSStores

### DIFF
--- a/resources/helm/dask-gateway/crds/traefik.yaml
+++ b/resources/helm/dask-gateway/crds/traefik.yaml
@@ -365,6 +365,89 @@ metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.4.1
   creationTimestamp: null
+  name: ingressrouteudps.traefik.containo.us
+spec:
+  group: traefik.containo.us
+  names:
+    kind: IngressRouteUDP
+    listKind: IngressRouteUDPList
+    plural: ingressrouteudps
+    singular: ingressrouteudp
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: IngressRouteUDP is an Ingress CRD specification.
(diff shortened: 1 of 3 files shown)


TomAugspurger:
I can't really comment on the changes here, but I'm happy to test things out (either on PRs if necessary, or prior to a release)

consideRatio:
@jcrist I just force pushed this with a rebase on master, fixing a trivial merge conflict in values.yaml

jcrist:
Have you tested this locally at all? It generally looks good, but our k8s tests may not be sufficient to test everything.

consideRatio:
Not more than the traefik instance and so actually starts running properly with these changes and version 2.5. I've not created dask-gateway clusters via k8s etc, so if that is missing from tests then it could be relevant to check manually as well which I haven't.

jcrist:
No, we do create clusters as part of the test. I'm mainly wondering if the traefik logs are filling up with errors or warnings or anything - sometimes traefik will complain about issues but still keep working. I'll need to do a full QA before release anyway to clean up some things, so I'll check this then. Fine to merge as is. Thanks!

consideRatio:
> I'm mainly wondering if the traefik logs are filling up with errors or warnings or anything - sometimes traefik will complain about issues but still keep working.

Ah, the logs didn't fill up with errors at least - I had also monitored the logs as I had to debug lots of errors before getting things functional to a running state.

---

Thanks for review and merge!!

PR #442, opened 2021-09-09
Avoid warning in Gateway.__del__

TomAugspurger:
This test was emitting a warning, which came from garbage collection on
Gateway objects that failed in `__init__`.

```pytb
E               Traceback (most recent call last):
E                 File "/home/taugspurger/src/dask/dask-gateway/dask-gateway/dask_gateway/client.py", line 376, in __del__
E                   not self.asynchronous
E                 File "/home/taugspurger/src/dask/dask-gateway/dask-gateway/dask_gateway/client.py", line 334, in asynchronous
E                   return self._asynchronous
E               AttributeError: 'Gateway' object has no attribute '_asynchronous'
```

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -372,7 +372,7 @@ class Gateway(object):
         self.close()
 
     def __del__(self):
-        if (
+        if hasattr(self, "_asynchronous") and (
             not self.asynchronous
             and hasattr(self, "_loop_runner")
             and not sys.is_finalizing()
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -372,7 +372,7 @@ class Gateway(object):
         self.close()
 
     def __del__(self):
-        if hasattr(self, "_asynchronous") and (
+        if hasattr(self, "asynchronous") and (
             not self.asynchronous
             and hasattr(self, "_loop_runner")
             and not sys.is_finalizing()


dask-gateway/dask_gateway/client.py
```diff
@@ -372,7 +372,7 @@ def __exit__(self, *args):
         self.close()
 
     def __del__(self):
-        if (
+        if hasattr(self, "_asynchronous") and (
```
jcrist, line 375: In case the property changes implementation, should maybe check the referenced name instead of an implementation detail.

```suggestion
        if hasattr(self, "asynchronous") and (
```

TomAugspurger: Let me check that. I assumed hasattr always returned true for properties, but maybe not if they raise an AttributeError?

TomAugspurger: OK switched to use `asynchronous` and there's no warning, but I have no idea why it's working. It seems like `hasattr` does always evaluate true in this case:

```python
In [1]: class Foo:
   ...:     @property
   ...:     def a(self):
   ...:         return self._a
   ...:

In [2]: hasattr(Foo, "a")
Out[2]: True
```

so I would expect the warning to be back when using the property name. But I guess it doesn't really matter, as long as it works.

jcrist: You're checking on the class above (which will always work), when what you meant to do was check on an instance

```python
>>> hasattr(Foo, "a")
True
>>> hasattr(Foo(), "a")
False
```

`hasattr` is basically:

```python
def hasattr(x, name):
    try:
        getattr(x, name)
        return True
    except AttributeError
        return False
```

jcrist:
Test failure is a known CI issue. Merging, thanks Tom!

PR #444, opened 2021-09-11
Helm chart: add note about dummy schema

### LINKED ISSUE #434: Helm chart: update the DaskCluster CRD to have a schema
Modern versions of CustomResourceDefinitions (CRDs) require one to specify a schema to for resources to follow. Currently, we have one, but it just sais anything goes. I suggest that we try to define this schema to strictly represent what is actually allowed instead.

### Related

https://github.com/dask/dask-gateway/blob/b288174ba347fbd911a503439f7d77c37a2ae19d/resources/helm/dask-gateway/crds/daskclusters.yaml#L20-L25

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/

### Action points

- Clarify what source code or similar can be used as a reference to create the schema.
- Consider if we can automatically generate the schema instead somehow, for example using [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder#readme)
- Create the schema

consideRatio:
Closes #434 by documenting we won't implement a schema for the CRD resources we define.

### DIFF
--- a/resources/helm/dask-gateway/crds/daskclusters.yaml
+++ b/resources/helm/dask-gateway/crds/daskclusters.yaml
@@ -18,10 +18,18 @@ spec:
       subresources:
         status: {}
       schema:
+        # NOTE: While we define a schema, it is a dummy schema that doesn't
+        #       validate anything. We just have it to comply with the schema of
+        #       a CustomResourceDefinition that requires it.
+        #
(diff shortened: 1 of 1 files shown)


jcrist (approved):
:+1:

PR #448, opened 2021-10-06
Use new Dask docs theme

jacobtomlinson:
Bumping the minimum docs theme to use the new version. This isn't strictly necessary but I wanted to be explicit here. This project doesn't use RTD so we don't have a CI preview, but I built it locally and things seem ok.

xref dask/dask-sphinx-theme#54


TomAugspurger:
Thanks!

PR #463, opened 2021-12-01
Fixed worker_threads config

TomAugspurger:
Restores the previous default of using the number of cores.

Closes #414 

Fixes #398

PR #464, opened 2021-12-02
Update docker images

TomAugspurger:
Updates all pinned packages in the Dockerfiles to their latest released versions.

I *haven't* updated Python from 3.8 to 3.9 or 3.10. Anyone have a preference on that? For now, I think we cut a release with Python 3.8 and then update to 3.9 shortly after the release.


TomAugspurger:
I'm going to rush this through, since https://github.com/dask/dask-gateway/pull/463 is failing without it. Will happily roll things back if these updated pins cause issues for anyone.

jcrist:
Thanks Tom!

PR #466, opened 2021-12-03
CI: Add automation to build/push docker images to ghcr.io (GitHub's container registry)

TomAugspurger:
This adds a GitHub action to build and push dask-gateway and dask-gateway-server to GitHub container registry on tags.

https://github.com/dask/dask-gateway/runs/4408165500?check_suite_focus=true has the logs from an earlier run that was set to run on all commits, not just tags (but not push). You can see the build output for dask-gateway at https://github.com/dask/dask-gateway/runs/4408165500?check_suite_focus=true#step:6:1.

### DIFF
--- a/.github/workflows/publish-images.yaml
+++ b/.github/workflows/publish-images.yaml
@@ -1,0 +1,83 @@
+name: Create and publish a Docker image
+
+# on:
+#   release:
+#     types:
+#       - created
+on:
+  push:
+    branches: [ "*" ]
+  pull_request:
+    branches: [ "*" ]
+
+
+env:
+  REGISTRY: ghcr.io
+
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image: [dask-gateway, dask-gateway-server]
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Get the version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
+          images: ${{ env.REGISTRY }}/${{ matrix.image }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
+        with:
+          context: ${{ matrix.image }}
+          # push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Set up Python 3.x
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Install release dependencies
+        run: |
+          python -m pip install --upgrade build twine
+
+      # - name: Build and publish kbatch
+      #   env:
+      #     TWINE_USERNAME: "__token__"
+      #     TWINE_PASSWORD: ${{ secrets.KBATCH_PYPI_PASSWORD }}
+      #   run: |
+      #     cd kbatch
+      #     python -m build
+      #     twine upload dist/*
+
+      # - name: Build and publish kbatch-proxy
+      #   env:
+      #     TWINE_USERNAME: "__token__"
+      #     TWINE_PASSWORD: ${{ secrets.KBATCH_PROXY_PYPI_PASSWORD }}
+      #   run: |
+      #     cd kbatch-proxy
+      #     python -m build
+      #     twine upload dist/*
+      #     cd ..
--- a/.github/workflows/publish-images.yaml
+++ b/.github/workflows/publish-images.yaml
@@ -1,19 +1,12 @@
 name: Create and publish a Docker image
 
-# on:
-#   release:
-#     types:
-#       - created
 on:
   push:
-    branches: [ "*" ]
-  pull_request:
-    branches: [ "*" ]
-
+    tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,48 @@
+name: Create and publish a Docker image
+
+on:
+  push:
+    tags:
+      - "*"
```
jacobtomlinson, line 6: Might want to constrain this down to `main`. Wouldn't want to open a feature branch and have images pushed as a result.

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Get the version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
```
jacobtomlinson, line 30: These are some hard pins. Perhaps it would be ok to relax this to `docker/login-action@v1`

jacobtomlinson (commented):
Thanks for opening this @TomAugspurger

.github/workflows/publish-images.yaml
consideRatio: This is constrained to tags being pushed!

.github/workflows/publish-images.yaml
consideRatio: Note I think dependabot translates these bumps of tags to hashes no matter what in the end if dependabot is configured to keep the feferenced github actions updated.

If it references a branch, its different though i think.

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
```
consideRatio, line 36: Can a comment be made to clarify what this does? Are you looking at the latest published image, and applies the same tags/labels to the new image?

.github/workflows/publish-images.yaml
jacobtomlinson: I guess I was just thinking I don't want this to run and send me a failure email when I push tags to my fork.

consideRatio: Hmm, are you ever pushing tags to your fork? I recognize the wish to avoid having a CI workflow run on my forks. I think even cronjobs are disabled by default on forks.

Can you clarify if you have a concrete change request @jacobtomlinson? I don't understand it clearly so I figure Tom may not do it either.

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,48 @@
...
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Get the version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
```
consideRatio, line 27: I understand this step to extract details about the version based on the pushed git tag that triggered this workflow. But, it is never used. If it was, we would see `steps.get_version.outputs.VERSION` be referenced somewhere.

.github/workflows/publish-images.yaml
consideRatio: > GitHub Action to extract metadata from Git reference and GitHub events.

Ah, so this is refining Git and GitHub metadata, so we can use it to tag/label our built docker images.

I'm not sure what git labels are extracted, and if it would make sense to set them on the built image as well - but it makes sense to extract versions from the git tag I think.

Given how this is configured, it will result in `v1.2.3` and `latest` tag if a tag `v1.2.3` is pushed. It can also be configured to emit tags `v1.2.3`, `v1.2`, and `latest` for example.

The use of this action, is the use of what the https://github.com/jupyterhub/action-major-minor-tag-calculator#readme does it seems.

consideRatio (commented):
I think this is a quite crucial contribution to get this repo's maintenance needs met.

@TomAugspurger I'd like to make a push towards getting this merged. Are you okay with me adding commits and leaving it to you, @jacobtomlinson, and other repo maintainers to take on the reviewer role?

TomAugspurger:
> @TomAugspurger I'd like to make a push towards getting this merged. Are you okay with me adding commits and leaving it to you, @jacobtomlinson, and other repo maintainers to take on the reviewer role?

Please do! Apologies for not getting back to this.

.github/workflows/publish-images.yaml
martindurant: I vote "resolved", unless it becomes a problem in the future.

.github/workflows/publish-images.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
dhirschfeld, line 34: One gotcha with using the github token is that it won't trigger dependent workflows - i.e. if you want a workflow to run when you publish an image that won't happen if it was published using the `GITHUB_TOKEN`:

image

* https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

In that case, the workaround is to use a PAT (usually stored as a repository or organisation secret). 

*Not sure if you were wanting to trigger other workflows, but if so hopefully that helps!*

jacobtomlinson: We have a bot user for this @dask-bot. There is an org secret called `DASK_BOT_TOKEN` that needs to be enabled for each repo that wants to use it, any Dask owner can do this in the org settings section. I am using it in the Dask Helm Chart because PRs raised by Actions do not trigger CI by default.

I've enabled it for this repo too so we can update this and it should be ok.

.github/workflows/publish-images.yaml
consideRatio: I'll address this via dependabot configuration to bump actions.

.github/workflows/publish-images.yaml
consideRatio: I needed to read up a lot about this action. Let's try it out!

PR #467, opened 2021-12-03
Added release notes

TomAugspurger:
I chose to name make this version `0.10.0`. Dunno if we want to move to date-based versioning like the other Dask projects.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,0 +1,25 @@
+Changelog
+=========
+
+0.10.0
+-------
+
+Released on December 3rd, 2021
+
+* Docker images: Update dependencies to the latest stable versions :pr:`464`
+* Docker images: some refactoring and arm64 support :pr:`423`
+* Fixed unintentional warning when cleaning up ``Gateway`` objects :pr:`442`
+* Fixed warning from use of the deprecated ``distributed.format_bytes`` :pr:`416`
+* Move CI to GitHub Actions :pr:`408`
+* Add support for authenticating with JupyterHub service :pr:`410`
+* Add warning about Go version to server installation instructions. :pr:`399`
+* Change default branch from master to main :pr:`372`
+* Add customizable worker_threads :pr:`353`
+* Add label to be allowed direct access to the jupyterhub pod :pr:`352`
+* Helm chart: update deprecated k8s resources no longer supported in k8s 1.22 :pr:`420`
+* Helm chart: refactor to use consistent modern syntax :pr:`425`
+* Helm chart: don't package a README.rst file :pr:`424`
+* Helm chart: update to traefik v2.5.x :pr:`431`
+* Helm chart: add imagePullSecrets for traefik :pr:`445`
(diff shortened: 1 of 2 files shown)


consideRatio:
Oh can we make it 1.0.0 instead? It helps communicate changes with semver. We had a alot of jupyterhub projects become 1.0.0 for this reason and it has been great - it becomes clear how to bump versions etc instead of unclear, for maintainers and users.

jcrist:
I'd actually rather use calver to match the other dask release conventions. dask-gateway isn't really released on a clear schedule, and the default images are just a snapshot of whatever dask/distributed/etc... releases were latest at that time. Calver seems like the right fit to me.

consideRatio:
Calver / semver both seem like great improvements from using 0.x.y as we can communicate so little with 0.x.y.

A big +1 for just going with one of calver / semver right now rather than sticking with 0.x.y.

Note that the helm chart version cant have leading zeroes like 20.04 in ubuntu, because it must be semver2 compliant in format according to helm3

martindurant:
@TomAugspurger , dd you want to update this to use calver?

TomAugspurger:
Fine by me. Is there a target date set, or do you want to set the date and version at https://github.com/dask/dask-gateway/pull/467/files#diff-ef86237172bb2f91c585e845b38d6fb28bb6ac128ff8efa541cdd86f81751bf2R4 and merge this in just prior to releasing?

martindurant:
It would be really nice to do it this month. There would be no need in the changelog, then, to specify the exact release date if it's already in the version.

martindurant:
In dask we are using YYYY.MM.X where X is 0, 1, 2 for as many releases as happen in a month. We will always be .0 :)

consideRatio:
@martindurant with `YYYY.MM.X`, we would have `2022.02.0` today if we cut a release - right? That would not be a valid SemVer2 version =/ The leading zero in the month makes it fail...

martindurant:
That's true, but pip tools strip the 0 anyway. If helm has a problem with it, .2. is fine too. This is for user interpretation mostly, and will still work with split-and-int either way.

consideRatio:
@martindurant okay so in this project - do we consistently go for YYYY.M.X format for the projects artifacts? We have the following artifacts influenced I think.

1. Notes in changelog etc
2. Git tags in this repo
3. conda-forge/PyPI packages
4. Docker images
5. Helm charts (must have YYYY.M.X rather than YYYY.MM.X)

martindurant:
Yes, that's OK with me. Perhaps let the other people on this thread have a say until end of today.

martindurant:
I suppose I forgot - still good to merge here?

consideRatio:
Merge and iterate sounds great to me.

I sidetracked a bit on a versioning discussion but it is my understanding that we are to go for a YYYY.M.X release for all of our release artifacts, which would be the Python packages, Helm chart and associated docker images.

martindurant:
+1 , we'll update the final version string when we have release credentials.

PR #474, opened 2022-02-11
Bump dask and distributed to 2022.02.0

consideRatio:


jrbourbeau:
FYI 2022.02.0 isn't on `conda-forge` yet, but should be in the next few hours

martindurant:
ready to test now? The format thing is still here, I see.

consideRatio:
I updated the comments to reference conda-forge versions instead of PyPI versions since we install from conda-forge.

I've rebased this PR on top of #477 which will make us not run into unrelated `black` formatting failures.

jacobtomlinson:
Woo pre-commit CI looks good here.

PR #477, opened 2022-02-12
ci: fix ci failure, optimize workflow triggers, document use of pre-commit.ci

### LINKED ISSUE #476: CI: avoid and fix ci failure, either by relying on pre-commit.ci or running pre-commit in a github workflow
### Issue summary

The `main` job in the `test` workflow [fail in a step](https://github.com/dask/dask-gateway/runs/5167437773?check_suite_focus=true) running `black . --check` because of the latest version of black is used, while it don't fail when using `pre-commit` which has a config with a pinned version. If we relied on pre-commit to run checks, instead of manually running `flake8` and `black` - the pinned versions would be used and this wouldn't happen.

### More than a solution to this

I'd like to propose a fix that does a bit ore than just fixing this.

In the jupyterhub ecosystem we started relying on running the pre-commit checks with the pre-commit.ci service which does three things - of which a github workflow would only easily do one.

1. It runs pre-commit.ci based on the .pre-commit-config.yaml file (very fast)
2. It automatically submit PRs if a version needs to be updated in a .pre-commit-config.yaml file
3. It automatically adds a commit to a PR with autoformatting changes if needed

An example all of of this in https://github.com/jupyterhub/jupyterhub/pull/3784 is when pre-commit.ci has:
- Created a PR to automatically update versions in a .pre-commit-config.yaml file
  image
- Automatically added a commit to a PR by having run pre-commit and letting it do autoformatting work (in this case, with new config)
  image
- Automatically run tests and reported results
  image

### Reference to related code

https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/.github/workflows/test.yaml#L46-L49

https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/.pre-commit-config.yaml#L1-L11

### Suggested action points

- [x] Activate pre-commit.ci as a service for this project.
   This is done by visiting the pre-commit.ci GitHub app and making it available for the dask GitHub organization to use: https://results.pre-commit.ci/
   And declare it to be used specifically on this github repository at: https://github.com/organizations/dask/settings/installations
   This means it would...
   - a) automatically runs a pre-commit CI/CD check like a github workflow would
   - b) automatically create PRs updating the version of pre-commit hooks defined in .pre-commit-config.yaml when new versions are out - at most once a week.
   - c) automatically add commits to PRs with autoformatting changes based on the .pre-commit-config.yaml file if needed
- [ ] Merge a PR, such as #477, removing the manual test step [here](https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/.github/workflows/test.yaml#L46-L49) and instead rely on pre-commit.ci to report if running pre-commit led to a failure.

consideRatio:
Ignore this PR until having considered #476!

If there is agreement on using the pre-commit.ci service as motivated within #476, then this PR closes #476.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,8 +1,23 @@
 name: Docs
 
 on:
+  pull_request:
+    paths:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/docs.yaml"
   push:
-    branches: ["main"]
+    paths:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/docs.yaml"
+    branches-ignore:
+      - "dependabot/**"
+      - "pre-commit-ci-update-config"
+    tags: ["**"]
+  workflow_dispatch:
 
 env:
   commit_msg: ${{ github.event.head_commit.message }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -56,11 +56,6 @@ jobs:
         run: |
           py.test tests/ -k 'not kubernetes' -v
 
-      - name: Flake8 and Black
-        run: |
-          flake8
-          black . --check
-
       - name: Go Tests
         run: |
           pushd dask-gateway-server/dask-gateway-proxy
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,11 +1,24 @@
 name: Test
 
 on:
-  push:
-    branches: [ "*" ]
   pull_request:
-    branches:
-      - main
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/*"
+      - "!.github/workflows/test.yaml"
+  push:
+    paths-ignore:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/*"
+      - "!.github/workflows/test.yaml"
+    branches-ignore:
+      - "dependabot/**"
+      - "pre-commit-ci-update-config"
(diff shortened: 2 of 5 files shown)


.github/workflows/docs.yaml
```diff
@@ -1,8 +1,25 @@
...
-    branches: ["main"]
+    paths:
+      - "docs/**"
+      - "**.md"
+      - "**.rst"
+      - ".github/workflows/*"
+      - "!.github/workflows/docs.yaml"
+    branches-ignore:
+      - "dependabot/**"
+      - "pre-commit-ci-update-config"
```
consideRatio, line 20: pre-commit.ci / dependabot pushes to these branches and opens a PR. So like this one would avoid triggering two workflow executions - one for the pushed branch, one for the PR.

martindurant:
I do like dormalising how things are done, and of course fast failures - all without worrying about upstream checker versions. Does this add any extra effort when developing locally?

consideRatio:
> Does this add any extra effort when developing locally?

Using pre-commit.ci would remove one piece of extra effort of developing locally and submitting a PR, which is to install and run `pre-commit`, which in turn run the `black` autoformatting hook as configured already in this repo. This is because pre-commit.ci would push a commit with the autoformatting changes needed to the PR if you had not done it yourself locally.

Whats critical and should be fixed no matter what, is that we now have a `.pre-commit-config.yaml` referencing an outdated version of black, and then we run tests with a new version of black in the github workflow.

jacobtomlinson (approved):

jacobtomlinson:
Test failures here are unrelated. Merging.

Issue #422, opened 2021-08-21
ci: fix intermittent failures of the main-tests and pbs-tests test


consideRatio:
As observed by @costrouc in https://github.com/dask/dask-gateway/pull/392#issuecomment-829615662 and me in #420, the PBS test is intermittently failing unrelated to changes in PRs.

Let's try fix that!

Also the main test seem intermittent.

![image](https://user-images.githubusercontent.com/3837114/130842037-40b5cdbb-e9d9-4702-b926-b665e1e5d7d4.png)

### Logs from pbs test

```
Run ./continuous_integration/docker/pbs/start.sh
++ dirname ./continuous_integration/docker/pbs/start.sh
+ this_dir=./continuous_integration/docker/pbs
++ cd ./continuous_integration/docker/pbs
++ pwd
+ full_path_this_dir=/home/runner/work/dask-gateway/dask-gateway/continuous_integration/docker/pbs
++ cd /home/runner/work/dask-gateway/dask-gateway/continuous_integration/docker/pbs/../../..
++ pwd
+ git_root=/home/runner/work/dask-gateway/dask-gateway
+ docker run --rm -d --name pbs -h pbs -v /home/runner/work/dask-gateway/dask-gateway:/working -p 8000:8000 -p 8786:8786 -p 8088:8088 --cap-add=SYS_RESOURCE daskgateway/testing-pbs
Unable to find image 'daskgateway/testing-pbs:latest' locally
latest: Pulling from daskgateway/testing-pbs
ac9208207ada: Pulling fs layer
e9270b8e0c2a: Pulling fs layer
404a529e534a: Pulling fs layer
8ebdb67ecf9e: Pulling fs layer
10da784d767c: Pulling fs layer
4c7ef137d5c7: Pulling fs layer
10da784d767c: Waiting
4c7ef137d5c7: Waiting
8ebdb67ecf9e: Waiting
e9270b8e0c2a: Verifying Checksum
e9270b8e0c2a: Download complete
8ebdb67ecf9e: Verifying Checksum
8ebdb67ecf9e: Download complete
ac9208207ada: Verifying Checksum
ac9208207ada: Download complete
4c7ef137d5c7: Verifying Checksum
4c7ef137d5c7: Download complete
404a529e534a: Verifying Checksum
404a529e534a: Download complete
ac9208207ada: Pull complete
10da784d767c: Verifying Checksum
10da784d767c: Download complete
e9270b8e0c2a: Pull complete
[...]

PR #479, opened 2022-02-16
Helm chart: update traefik CRDs and Traefik version from 2.5 to 2.6

consideRatio:
The Helm chart relies on Traefik, and Traefik relies on some CRDs be pre-installed, so we bundle those with the dask-gateway helm chart. This installation updates those bundled CRDs and the referenced Traefik version along with them.

### DIFF
--- a/resources/helm/dask-gateway/crds/traefik.yaml
+++ b/resources/helm/dask-gateway/crds/traefik.yaml
@@ -1,7 +1,9 @@
 # Don't manually update these CRDs, if needed instead re-import them from
-# https://github.com/traefik/traefik-helm-chart/tree/HEAD/traefik/crds.
+# https://github.com/traefik/traefik-helm-chart/tree/HEAD/traefik/crds and
+# consider the changes since last made via commit history:
+# https://github.com/traefik/traefik-helm-chart/commits/HEAD/traefik/crds
 #
-# The most recent import was made from commit 3a33968.
+# The most recent import was made from commit 693589e.
 #
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -741,6 +743,7 @@ spec:
                       type: string
                     type: object
                   featurePolicy:
+                    description: 'Deprecated: use PermissionsPolicy instead.'
                     type: string
                   forceSTSHeader:
                     type: boolean
@@ -752,6 +755,8 @@ spec:
                     type: array
                   isDevelopment:
                     type: boolean
+                  permissionsPolicy:
+                    type: string
                   publicKey:
                     type: string
                   referrerPolicy:
@@ -1040,6 +1045,13 @@ spec:
           spec:
             description: MiddlewareTCPSpec holds the MiddlewareTCP configuration.
             properties:
+              inFlightConn:
+                description: TCPInFlightConn holds the TCP in flight connection configuration.
+                properties:
+                  amount:
+                    format: int64
+                    type: integer
+                type: object
               ipWhiteList:
                 description: TCPIPWhiteList holds the TCP ip white list configuration.
                 properties:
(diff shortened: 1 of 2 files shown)


martindurant:
+1, I'll test later this week.
No idea why PBS is failing, but the main test is only failing on formatting ( #477 #476 )

dhirschfeld:
Just confirming that the bundled CRDs won't interfere if I have an AKS cluster with `traefik` already deployed?
```
> kubectl get crd -A | rg traefik
ingressroutes.traefik.containo.us                                  2022-02-14T02:36:48Z
ingressroutetcps.traefik.containo.us                               2022-02-14T02:36:48Z
ingressrouteudps.traefik.containo.us                               2022-02-14T02:36:48Z
middlewares.traefik.containo.us                                    2022-02-14T02:36:48Z
middlewaretcps.traefik.containo.us                                 2022-02-14T02:36:48Z
serverstransports.traefik.containo.us                              2022-02-14T02:36:48Z
tlsoptions.traefik.containo.us                                     2022-02-14T02:36:48Z
tlsstores.traefik.containo.us                                      2022-02-14T02:36:48Z
traefikservices.traefik.containo.us                                2022-02-14T02:36:48Z
```

consideRatio:
@martindurant I'm not sure about the PBS failure's root cause - but I'm quite confident it's unrelated. I opened this issue about it: https://github.com/dask/dask-gateway/issues/422

@dhirschfeld the dask-gateway helm chart does indeed install some CRDs, and I'm not sure how `helm` handles it if there are already CRDs installed or similar. I'm quite confident though that this PRs change wouldn't break anything by keeping them updated.

For details on how `helm` handles a helm chart with a `crds` folder, [there is this documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)

> These CRDs are not templated, but will be installed by default when running a helm install for the chart. **If the CRD already exists, it will be skipped with a warning.**

> There is no support at this time for upgrading or deleting CRDs using Helm.

In practice, this means that a fresh install of dask-gateway will provide fresh CRDs, and if you upgrade after this has been made - I believe nothing will change in practice: the old CRDs will be used. This **could be** a problem that needs to be managed by communicating to the end users that the dask-gateway helm chart relies on modern traefik CRDs for its modern traefik version - and because of that, will need to manually delete the traefik CRDs and re-run a `helm upgrade` for example.

A discussion about this in depth is probably not relevant for the scope of this PR though - I suggest opening an issue if you want to deliberate on it further. So far, we haven't had problems with this.

---

I've rebased this on the main branch that includes some CI fixes, hopefully the PBS test doesn't intermittently crash on us again.

dhirschfeld:
Thanks for the info @consideRatio! I imagine if you were trying to run two very different versions there *might* be some issues but I'm using close-to-the-latest version and I don't think the CRDs change too much so it should be fine.

Something to keep in mind and if I ever do run into problems I can open an issue with a concrete example.

martindurant:
I can confirm that the updated traefik and CRDs work fine (and with latest dask too)

martindurant:
As a follow up, we might consider documentation about which versions of what we support; for example, the new CRD versions backport for a while, but how long? I suspect finding exact version bounds on things may be quite an involved process.

consideRatio:
@martindurant yeah it's tricky, especially since the versions of the CRDs are not really changed - they modify an existing version of the k8s resource they call v1alpha1.

I think what we should do, is to realize via helm chart upgrade test, if a change will break an existing installation when upgrading. Then, if we conclude that it did - then we can make a decision on if we can/should do something to be backward compatible.

I'm quite lost in general on the issues until they are demonstrated practically - so having an upgrade test of a helm chart installation seems like step 1. I opened https://github.com/dask/dask-gateway/issues/480 to reflect that need.

Thanks @martindurant and @dhirschfeld for your review efforts!

PR #485, opened 2022-03-01
Remove no longer used (?) github repo deploy key (to push to gh-pages branch?)

consideRatio:
I'm not sure, but I think this repo has a deployment key that has been used when used together with TravisCI as a means for TravisCI to push back updates to the `gh-pages` branch of this repo.

It was added as part of https://github.com/dask/dask-gateway/pull/107, I think perhaps it was outdated even at that time as I found no reference to the filename in the repo, and I saw that the TravisCI had an environment variable that probably was used to help provide credentials to a utility pushing built docs to the gh-pages branch.

If it is a deployment key for this repo, as indicated by the filename `github_deploy_key_dask_dask_gateway.enc`, it should be registered at https://github.com/dask/dask-gateway/settings/keys. It should reasonably be cleared from there as well.


martindurant:
I can confirm that there are no registered keys for this repo.

jcrist:
This is left over from using doctr to publish docs on travis ci, perfectly fine to delete it now.

PR #492, opened 2022-03-20
ci: misc updates to test workflow and pytest-asyncio

consideRatio:
- ci: test workflow, readability refactoring
- ci: test workflow, update versions
- ci: try unpin pytest-asyncio
- ci: asyncio_mode=auto makes us not need to decorate async tests

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,17 +31,17 @@ jobs:
   main-tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v3
         with:
           python-version: "3.8"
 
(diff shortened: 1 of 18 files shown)


.github/workflows/test.yaml
```diff
@@ -76,16 +76,18 @@ jobs:
...
-      - uses: AbsaOSS/k3d-action@v1.5.0
+      - uses: actions/checkout@v3
+
+      - uses: AbsaOSS/k3d-action@v2
         name: "Create k3d Cluster"
         with:
           cluster-name: "k3s-default"
           args: >-
-            -p "30200:30200@agent[0]"
+            --port "30200:30200@agent:0:direct"
```
consideRatio, line 86: This related to breaking changes in bumping the k3d action, which now used k3d v5 instead of k3d v4.

jacobtomlinson (approved):
This looks great thanks @consideRatio. We should probably think about setting `pytest-asyncio` to `auto` in other projects too.

PR #494, opened 2022-03-20
ci: build/push python packages workflow, added

consideRatio:


jacobtomlinson (commented):
This looks great. I'll leave this until we get unblocked on the credentials.

jcrist:
Thanks Jacob and Erik, what are y'all's pypi usernames? I'll add you right now.

martindurant:
@jcrist , that's great news. Please also update the conda-forge feedstock.

jacobtomlinson:
Awesome thanks @jcrist. I am `jacob.tomlinson` and Erik is `consideRatio`. If you could merge conda-forge/dask-gateway-feedstock#22 also that would be great!

jcrist:
I'm not sure what your wheel building automation plans are, but note that dask-gateway-server has some native (go) code that will need to be fixed and checked by auditwheel. Previously I had a hacky buildscript I'd run locally for the osx & linux builds, but if I were to redo this today I'd just use cibuildwheel and be done with it. Apologies if y'all already know all this.

consideRatio:
@jcrist thank you for clarifying this and suggesting a strategy to go about it!

I have much to learn about Go but this seems like a good reason to learn some more. I'll rework this GitHub workflow as needed to make things work as they should.

Jim Is the hacky buildscript you had version controlled in this repo or something you could share in a gist or similar?

jcrist:
The scripts are here: https://github.com/dask/dask-gateway/tree/main/dask-gateway-server/build_tools, but things should build fine with cibuildwheel, provided the golang compiler is added to the build images before running the setup.py. This could be done with https://cibuildwheel.readthedocs.io/en/stable/options/#before-all perhaps. Note that `dask-gateway-server` doesn't support windows, so a bash script for linux/osx should be sufficient.

consideRatio:
Ah excellent, I'm confident I can come up with something sensible with this very helpful guidance!

I opened https://github.com/dask/dask-gateway/issues/496 to track the need to fix this properly, something I'd be happy to do.

PR #497, opened 2022-03-24
ci: let flake8 be configured in a single place instead of three

consideRatio:
I saw three configuration locations for flake8, each of which had the same configuration. Since pre-commit will be run and register the `.flake8` configuration file in the git repo's root folder, I figure it made sense to avoid having multiple dupicates of this config.

I also saw that the central .flake8 configuration mentioned two not existing files that wasn't .gitignored either, I assume they have been part of the repo but no longer are. Due to that, I figured it made sense to remove them as well.


.flake8
```diff
@@ -1,9 +1,7 @@
 [flake8]
 exclude =
     __init__.py,
-    versioneer.py,
-    _version.py,
-    dask_gateway_config.py
```
consideRatio, line 6: - versioneer.py isn't used in this project
- dask_gateway_config.py may be generated, but is .gitignored as part of #499

Issue #498, opened 2022-03-24
Drop support for python 3.7


consideRatio:
At this point Python 3.6 has reached end of life. If we wish to support it, we may need to not use modern versions of libraries etc as they may no longer support Python 3.6.

Due to that, given agreement, I suggest we drop support for Python 3.6 at this point if there is agreement about it.

jacobtomlinson:
We only support `3.8` and above. I would suggest we should be consistent here too.

dask/community#213

consideRatio:
In #501 I've now added a commit to drop Python 3.6 and require Python 3.7+.

To drop support for Python 3.7 without breaking our test suite etc, some additional work needs to be done.

### Action points
- [ ] Drop support for Python 3.6 (#501)
- [ ] Drop support for Python 3.7

PR #501, opened 2022-03-25
maint/ci: remove support for Python 3.6 and test against multiple versions of Python and Golang

consideRatio:
I wanted to better understand if this Python code functioned in Python 3.6-3.10, as well as if its associated Golang code in dask-gateway-server/dask-gateway-proxy would function to compile with Golang 1.12-1.18.

I couldn't arrive at a clear conclusion about Golang, but we seem to be fine with Python 3.6-3.10. A key reason that we don't properly test against all versions of Golang is that we have golang versions in many different contexts. For example, we have it in the Dockerfile's defined to start slurm/pbs/hadoop to test against them.

I raised the question if we should drop support for Python 3.6 which has reached end of life three months ago in #498. I think it makes sense to do so at this point in time.

---

This PR is also updated to remove support for Python 3.6 which has reached end of life.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,16 +30,61 @@ env:
 jobs:
   main-tests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # We test against different versions of Python and Golang, but not
+          # against different versions of Node.
+          #
+          # - The Python version installed where dask-gateway and
+          #   dask-gateway-proxy is run by the end user matters, so we test
+          #   against all the versions we intend to support.
+          #
+          #   We could for example choose to test against the versions that
+          #   hasn't reached end of life yet: https://endoflife.date/python.
+          #
+          # - The Golang version that compiles
+          #   dask-gateway-server/dask-gateway-proxy bundled for the
+          #   dask-gateway-server Python package is the only thing that matters.
+          #   Due to that, we can test fewer versions of Golang.
+          #
+          #   We could for example choose to test against the versions that
+          #   hasn't reached end of life yet: https://endoflife.date/go.
+          #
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -30,16 +30,58 @@ env:
...
+          #   FIXME: Golang 1.16 and above fails because our "go get" isn't
+          #          functioning as before. Golang 1.12 and below fails because
+          #          the stretchr/testify/assert library doesn't support golang
+          #          1.12 and below.
+          #
+          # - Node is a dependency for JupyterHub's configurable-http-proxy that
+          #   we test integration with. We can test against only one version and
+          #   that would be fine.
+          #
+          - python-version: "3.6"
```
jacobtomlinson, line 64: Dask as a whole dropped `3.6` recently, so I suggest we don't worry about supporting it here.

consideRatio: Oh, okay hmmmm let's consider dropping support for Python 3.7 at a later time. For now, I added a commit that stops tests against Python 3.6 and requires Python 3.7+. Various tests rely on pre-build Dockerfiles that still use Python 3.7 etc so it would be  big bloat to make such change as part of this PR.

jacobtomlinson (approved):
This looks good to me. Happy for merge after passing CI (GitHub Actions is down for maintenance right now).

consideRatio:
Rebased on main and removed a setup.py comment about Python 3.6.

```
Classifier: Programming Language :: Python :: 3.6
```

consideRatio:
As this was approved for merge after CI passed by @jacobtomlinson in https://github.com/dask/dask-gateway/pull/501#pullrequestreview-921824673, and that I've only removed a outdated comment about Python 3.6 since, I'll go for a self-merge on this to be able to avoid conflicts with future development effort on dropping Python 3.7.

PR #502, opened 2022-03-25
Remove legacy purge flag from k8s uninstall docs

brews:
Minor doc fix!

The [install instructions for dask-gateway on Kubernetes](https://gateway.dask.org/install-kube.html#shutting-everything-down) use an unsupported flag (`--purge`) in the `helm` command for cleanup and uninstalling work. When run, this throws an error at the poor user.

This updates the docs to remove the `--purge` flag. No other change is needed because behavior from `--purge` is the default behavior in the current version of `helm`.

### DIFF
--- a/docs/source/install-kube.rst
+++ b/docs/source/install-kube.rst
@@ -184,7 +184,7 @@ clean everything up. You can do this with ``helm delete``:
 
 .. code-block:: shell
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Thank yo @brews!!

PR #505, opened 2022-03-26
ci: run go native tests against modern versions of go

consideRatio:
I didn't understand enough about go to successfully run the go native tests defined in dask-gateway-server/dask-gateway-proxy against versions 1.16+. I've now pieced together some additional understanding to get this to work. I think the gist was that we had to install golang test dependencies part of the project, and not globally for Go 1.16+.

This could perhaps be worked around by having a global environment variable GO111MODULE=auto, or setup the test dependencies in another way somehow? I'm not sure. For now, this change to run `go get` from the folder with the golang code seems like a good resolution that seem to work for many versions of golang.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -52,26 +52,18 @@ jobs:
           #   We could for example choose to test against the versions that
           #   hasn't reached end of life yet: https://endoflife.date/go.
           #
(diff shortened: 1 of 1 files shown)


jcrist (approved):
Makes sense to me, thanks!

consideRatio:
Thank you @jcrist for review/merge!!

PR #509, opened 2022-03-26
Add _ prefix to "internal use" functions in controller.py

consideRatio:
I saw a few functions I figured out was "internal use" functions not relevant outside the file and added a leading prefix underscore to them. [PEP8 describes naming conventions](https://peps.python.org/pep-0008/#descriptive-naming-styles), and here is a quote on using a leading underscore:

> `_single_leading_underscore`: weak “internal use” indicator. E.g. from M import * does not import objects whose names start with an underscore.

I considered if I should make a PR about this detail at all, and now did and ended up learning more about PEP8! I mainly valued that by using a underscore prefix, it became clear for anyone reading the code that these are minor helper functions not being used elsewhere.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
@@ -39,7 +39,7 @@ from .utils import (
 from .backend import KubeBackendAndControllerMixin
 
 
-def get_container_status(pod, name):
+def _get_container_status(pod, name):
     """Get the status of container ``name`` from a pod"""
     for cs in pod["status"].get("containerStatuses", ()):
         if cs["name"] == name:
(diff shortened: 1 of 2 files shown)


jcrist:
I'm -0.5 on this change. While I agree that a `_` prefix is good to indicate internal methods/classes/functions in public modules (things that a user will import), all of `dask-gateway-server` (minus `dask_gateway_server.options` and the publically documented config) is considered private. It's an application, not a library. Adding a `_` prefix everywhere would muddy the code, and doing it only in a few places makes this a judgement call rather than a clear rule. Since this code currently works fine and no user has run into an issue where they're trying to dig into the internals, it doesn't make sense to me to focus on cosmetic changes like this.

consideRatio:
Thanks for your review, that makes sense to me to @jcrist!

PR #510, opened 2022-03-26
Cleanup no longer needed workarounds for Python 3.6

consideRatio:
With #501 support for Python 3.6 was dropped, so can assume Python 3.7+. This PR removes checks for `Python version >= 3.7` which will now always be true.


jcrist (approved):
Thanks @consideRatio!

Issue #511, opened 2022-03-26
ci: kubernetes test intermittently hangs


consideRatio:
As can be seen in [this job](https://github.com/dask/dask-gateway/runs/5703554767?check_suite_focus=true) the Kubernetes test can hang on the step named Helm install:

https://github.com/dask/dask-gateway/blob/b23301a6923bcb3e0651519b46c3a8d8bc2cba71/.github/workflows/test.yaml#L143-L148

![image](https://user-images.githubusercontent.com/3837114/160254081-e3097930-8af9-4697-a7d1-07ddb9f78d96.png)

Specifically, it seems that importing the built image to be available for the k8s cluster's container runtime is what can hang.

https://github.com/dask/dask-gateway/blob/b23301a6923bcb3e0651519b46c3a8d8bc2cba71/continuous_integration/kubernetes/helm-install.sh#L21-L25

## Workaround ideas

- [x] We can add a timeout to avoid this hanging for 6 hours which is the maximum time for a github workflow's job.
  https://github.com/dask/dask-gateway/pull/512
- [ ] We can avoid the need to import the images by using k3s instead of k3d, by using the [jupyterhub/action-k3s-helm](https://github.com/jupyterhub/action-k3s-helm#readme) GitHub Action to setup a k3s cluster.
  #518

I think we should do both, starting with 1 as it is such a primitive quick and easy thing to reduce the impact of this issue. I searched for `k3d` and found that there is a undocumented and unreferenced `k3d-create.sh` script. I think this is good to retain this script and document it, because when setting up a local development environment, k3s is less suitable as it would be more intrusive to install locally. `k3d` is a wrapper for `k3s` to run in docker, makes most sense to use on your local computer for testing, while `k3s` can be used without trouble in a CI environment.

jcrist:
Hmmm, that's odd that it hangs sometimes now, I don't remember ever seeing that issue before. I personally have no strong attachment to the current k8s CI setup - k3d was the easiest option at the time I wrote this, if there's a more robust way now I'd be happy to move to it. Thanks for opening this Erik!

PR #512, opened 2022-03-26
ci: add timeout to avoid 6h consequence of intermittent hang issue

consideRatio:
Mitigates but does not close_ #511.


jcrist (approved):
:+1:

PR #517, opened 2022-03-27
ci: add fixme notes, update python/go versions, make a script executable like others

consideRatio:


.github/workflows/build-publish-docs.yaml
```diff
@@ -42,20 +42,36 @@ jobs:
...
       - uses: actions/setup-python@v3
         with:
-          python-version: "3.8"
-
+          python-version: "3.10"
+
+      # FIXME: go is used to compile dask-gateway-server/dask-gateway-proxy, but
+      #        that isn't relevant for us when building documentation, but, we
+      #        end up doing it anyhow currently because we re-use a course
+      #        installation script.
```
jcrist, line 50: Note that the setup.py script supports a --no-build-proxy option to skip compiling the go proxy. That could be used for the docs builds to avoid this dependency, but I don't think that's super high priority. Installing go in ci is pretty quick.

jcrist (approved):

PR #518, opened 2022-03-27
ci: use k3s instead of k3d to setup k8s, and test against k8s 1.20-1.23

### LINKED ISSUE #511: ci: kubernetes test intermittently hangs

consideRatio:
By using k3s in our CI setup, we can avoid needing to load images into the k8s container runtime as we would need to if we used k3d. This saves us a minute or so for each kubernetes test we run. It seems that the average time doesn't improve though because it seems using k3d was a bit faster to get started. The k3s the jupyterhub/action-k3s-helm includes `calico` to help enforce NetworkPolicy resources in a reliable way which the k3d setup didn't provide.

By using the jupyterhub organizations GitHub Action to setup k3s we also get kubectl/helm setup for us. Combied with use of the jupyterhub organizations GitHub Action to provide a report of what has happened within the namespace, we can remove the install-tools.sh script installing kubectl, helm, and stern used to capture misc logs.

Closes #511


jcrist (approved):
Thanks Erik! Always nice to delete code.

PR #519, opened 2022-03-27
ci: de-duplicate deps by docs/requirements.txt and tests/requirements.txt and update CI images

### LINKED ISSUE #401: paywalling of cloudera repo
The CI dockerfile for Yarn fails with 404-unautheriized. This appears to be because of the paywalling by Cloudera described in:
https://www.cloudera.com/downloads/paywall-expansion.html

### LINKED ISSUE #516: De-duplicate package dependencies
I think there is an unsustainable maintenance related to the amount of places we have dependencies listed for dask-gateway and dask-gateway-server and various test dependencies.

As an example, this is where `aiohttp` is mentioned as part of being listed as a dependency.

- [dev-environment.yaml](https://github.com/dask/dask-gateway/blob/main/dev-environment.yaml)
- [continuous_integration/install.sh](https://github.com/dask/dask-gateway/blob/HEAD/continuous_integration/install.sh)
- [continuous_integration/kubernetes/install.sh](https://github.com/dask/dask-gateway/blob/HEAD/continuous_integration/kubernetes/install.sh)
- [continuous_integration/docker/hadoop/install.sh](https://github.com/dask/dask-gateway/blob/HEAD/continuous_integration/docker/hadoop/install.sh)
- [continuous_integration/docker/pbs/install.sh](https://github.com/dask/dask-gateway/blob/HEAD/continuous_integration/docker/pbs/install.sh)
- [continuous_integration/docker/slurm/install.sh](https://github.com/dask/dask-gateway/blob/HEAD/continuous_integration/docker/slurm/install.sh)
- [dask-gateway/Dockerfile](https://github.com/dask/dask-gateway/blob/HEAD/dask-gateway/Dockerfile)
- [dask-gateway/setup.py](https://github.com/dask/dask-gateway/blob/HEAD/dask-gateway/setup.py)
- [dask-gateway-server/Dockerfile](https://github.com/dask/dask-gateway/blob/HEAD/dask-gateway-server/Dockerfile)
- [dask-gateway-server/setup.py](https://github.com/dask/dask-gateway/blob/HEAD/dask-gateway-server/setup.py)

## Ideas

- We decouple setup.py from listing core dependencies to a requirements.txt file
- We add a dev-requirements.txt file includes all `extra_require` dependencies in setup.py
- We make our Dockerfiles install the relevant requirements file or to make it install the dependencies automatically as part of installing the package somehow, as `pip install` would manage. Right now it sais `setup.py develop` and I'm not sure what that means but I know that [dask-gateway-server relies on that to handle a `no-build-proxy` option of some kind](https://github.com/dask/dask-gateway/blob/73a0182f6f5ab8601d42cece7ae7b694d070ea2e/dask-gateway-server/setup.py#L86-L88)

consideRatio:
## Important

While this is a big PR, it is about CI - it only touches two `setup.py` files which is directly related to end use. Overall, this is reducing complexity but since I've added a lot of comments and documentation there are a lot of additional lines.

image

---

## Summary

- Closes #516
  I thought it was a maintenance problem that there were so many of the same dependencies declared in multiple locations. I got confused by that overall and decided to address it before digging into #514.
  
  I've created docs/requirements.txt and tests/requirements.txt, that alongside dependencies declared in setup.py files makes for the only places we declare Python dependencies.
- Closes #401

  I wanted to rebuild the images that no longer included duplicated deps so I took on #401 here as well which was a major undertaking. We now test against Hadoop 3 / Yarn 2 instead of Hadoop 2 and Yarn 1.
- This PR unblocks work on #498 that would otherwise break our Yarn/Slurm/PBS tests. This is because if we start require a newer version than Python 3.7, we would still be stuck with our old images we run tests in for Yarn/Slurm/PBS which had a Python 3.7 version.
- This PR helped me debug #522 and create #525 as I could update `click` pinnings in fewer places and not miss a place which would have made me draw the wrong conclusions as I did initially.

---

## Intermittent failures

This PR initially introduced intermittent failures for the hadoop test, but they are now gone by restricting memory settings on scheduler/workers started during tests. See #422 for details on already existing intermittent test failures though.


martindurant:
> which was a major undertaking.

:)  
I will look it over, not sure how long it will take or how authoritative I can be!

consideRatio:
@martindurant I think I figured it out, it was a newline character in CLASSPATH which made the last entry fail to function properly. Figuring this out took me a full day of effort ;D

consideRatio:
@martindurant @jacobtomlinson @jcrist this is ready for review now. I've updated the original PR description to be more succinct about the changes.

consideRatio:
@jcrist I've verified that the wheel with skein 0.8.2 that was recently released functions as it should as part of this PR!

consideRatio:
@martindurant @jacobtomlinson @jcrist I'd love to get this merged and iterate from there, this includes too many changes as it is.

Next work item is to make tests more reliable, the main tests, the pbs tests, and now also the hadoop tests. I don't know what makes them be unreliable though besides guesses that it at least partially involves a OOMKiller sending SIGKILL causing exit code 137 for the PBS test, which then shows symptoms of "ValueError, 404 NOT FOUND" etc, which is what I also see in main/hadoop test.

continuous_integration/docker/hadoop/_print_logs.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -x
```
martindurant, line 2: When does this one get run?

I would include the name of each log file before its contents.

continuous_integration/docker/hadoop/_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -xe
+
+cd /working
+pytest -v \
+    tests/test_yarn_backend.py \
```
martindurant, line 6: Silly question, but if we explicitly pick out yarn tests here, is there any point in the YARN_TEST (or whatever it's called) env variable? Should we set it here?

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/core-site.xml
```diff
@@ -45,24 +45,4 @@
         <value>org.apache.hadoop.security.AuthenticationFilterInitializer</value>
     </property>
 
-    <property>
-        <name>hadoop.http.authentication.type</name>
```
martindurant, line 49: You disabled HTTP auth altogether? I agree with that.

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/mapred-site.xml
```diff
@@ -5,6 +5,7 @@
         <value>yarn</value>
     </property>
 
+    <!-- networking -->
```
martindurant, line 8: (I'm pretty sure we don't need this file at all)

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/yarn-site.xml
```diff
@@ -61,18 +67,19 @@
     </property>
 
     <property>
-        <name>yarn.nodemanager.resource.cpu-vcores</name>
-        <value>16</value>
+        <name>yarn.nodemanager.resource.memory-mb</name>
+        <value>4096</value>
```
martindurant, line 71: You probably know, but YARN doesn't actually respect these as limits, it's only used for internal bookkeeping; so it doesn't matter is the worker actually had this much memory available.

martindurant (commented):
I have looked through much of the YARN stuff, and I have very little problem with anything, it all looks solid. Of course, there are loads of configs, so I could easily have missed something.

continuous_integration/docker/base/Dockerfile
```diff
@@ -1,33 +1,79 @@
-FROM centos:7.6.1810
+# See continuous_integration/docker/README.md for details about this and other
+# Dockerfiles under the continuous_integration/docker folder on their purpose
+# and how to work with them.
+#
+# centos:8 reached end-of-life 31 Dec 2021
+# centos:7 reach end-of-life 30 Jun 2024
```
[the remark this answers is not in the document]
consideRatio: I figure it is like ubuntu 20.04 vs 21.10, one is LTS, one isnt. Do you choose the latest and get weaker software support or the older with better?

I wanted to be compatible with software without needing to re-compile it, centos 7 vs centos 8 comes with ups and downs. Various versions with PBS supported various versions of centos etc. Since centos 8 has reached end of life already, i heaitated to bump to that as it probably would fail to be compatible with software going onwards even if it didnt yet.

There are even more modern versions i think, but overall this seems reasonable to stick with to reduce complexity and avoid work.

continuous_integration/docker/hadoop/_print_logs.sh
consideRatio: > When does this one get run?

This is from test.yaml, a github workflow. It is the last line.

```yaml
      - name: Print logs for debugging
        if: always()
        run: ./continuous_integration/docker/hadoop/print_logs.sh
```

continuous_integration/docker/hadoop/_test.sh
consideRatio: That is a good point, I agree that we should instead rely on that instead, but it is a non-critical change so I may not drive making these tests skipped by default unless marked for execution - which could be what would be the cleanest strategy.

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/core-site.xml
consideRatio: I saw it failing anyhow, so I figured removing it reduced complexity without downside.

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/mapred-site.xml
consideRatio: Figured as well, didn't dare go test if that would be ok or not but then I'll go for removal of it!

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/yarn-site.xml
consideRatio: Ooh, that is critical knowledge. I wanted to stay within memory limits and constrain this to avoid crashing things during tests. I had seen a warning about possible "thrashing" of workers or similar.

Ideas on how to make us limit ourselves better?

consideRatio: Is this config also pointless?

```xml
    <property>
        <name>yarn.scheduler.minimum-allocation-mb</name>
        <value>32</value>
    </property>

    <property>
        <name>yarn.resource-types.memory-mb.increment-allocation</name>
        <value>${yarn.scheduler.minimum-allocation-mb}</value>
    </property>
```

martindurant: I did get tripped up by this once, where where the default total memory resource of the cluster was 4096 and the minimum allocation was 1024, so you would get no more than 4 containers no matter what the actual memory requirements of requests were. The default minimum still seems to be 1024 https://hadoop.apache.org/docs/r3.0.1/hadoop-yarn/hadoop-yarn-common/yarn-default.xml 

I would make these numbers small, or else just wait in case something ever goes wrong.

consideRatio: @martindurant I found that in the test_yarn_backend.py file, the YarnBackend is configured with memory for the workers. I lowered it from 512 to 128 and things may have improved, not sure yet - intermittency =/

continuous_integration/docker/hadoop/_print_logs.sh
consideRatio: > I would include the name of each log file before its contents.

Do you mean that you would do an echo manually before each line? That is reasonable, and perhaps more reliable - I think `set -x` has led to some weird timings where some logs show up before/after etc.

consideRatio:
@martindurant I think this has become a lot more robust now! I think the PBS and hadoop tests were both memory related, and now they may be stable. But yeah... I may be wrong. I have been wrong so many times.

continuous_integration/docker/hadoop/_test.sh
consideRatio: I'm opting to not go for this refactoring, it can absolutely be done and I'll +1 for merging a PR about this, but for now I'll avoid doing that refactoring.

continuous_integration/docker/hadoop/_print_logs.sh
consideRatio: I removed `set -x` and added colorized print statements on what commands are run.

consideRatio:
@martindurant thanks for review efforts!!! :tada: :heart: - to know you were around giving it a look made a big difference on my motivation to get this done!

continuous_integration/docker/hadoop/files/etc/hadoop/conf.kerberos/core-site.xml
consideRatio: I'm confused. This is the default value anyhow so this didn't change much I think, but what does this config really do? I think it influences something that we may not be using?

martindurant:
I'm not sure when I'll have the time to look through again :|  If you think things are in a good state, I trust you!

consideRatio:
> I'm not sure when I'll have the time to look through again :| If you think things are in a good state, I trust you!

Thank you! I think this is good to go then!

Issue #522, opened 2022-04-05
CI failures for main/k8s tests has been introduced without changes in this repo


consideRatio:
In [this old test run](https://github.com/dask/dask-gateway/runs/5709653936?check_suite_focus=true) of a recent commit, all tests passed. In [this newer test run](https://github.com/dask/dask-gateway/runs/5808760416?check_suite_focus=true) _of the same commit_ (manually triggered) we now have failures in the Python 3.7 test and all the k8s tests.

Something changed outside this repo, but what?

## Pip list differences

The previous worked, the new didn't. but something else can also be the cause.

```diff
-click==8.0.4
+click==8.1.2
-fsspec==2022.2.0
+fsspec==2022.3.0
-ipykernel==6.9.2
+ipykernel==6.11.0
-jupyter-client==7.1.2
+jupyter-client==7.2.1
-nbconvert==6.4.4
+nbconvert==6.4.5
-nest-asyncio==1.5.4
+nest-asyncio==1.5.5
-SQLAlchemy==1.4.32
+SQLAlchemy==1.4.34
-zipp==3.7.0
+zipp==3.8.0
```

martindurant:
I did see `click` implicated in some breakages elsewhere, but I don't remember quite where.

consideRatio:
I narrowed it down, this is worked around by avoiding click 8.1.0, 8.1.1, and 8.1.2 (latest), but I couldn't figure out what changed in click that caused this.

consideRatio:
Sadly this did only resolve the issue for the py37 test, not for all the k8s tests. I don't get whats the problem there still, but have started suspecting memory issues.

consideRatio:
Ah... I realized now that `click` probably is too blame also in the k8s tests. It just was that the `setup.py` defined version constraint wasn't respected by our Dockerfile's. I opened #524 about this.

consideRatio:
I'll let this be closed by #525 and open a new about fixing compatebility with click 8.1.0+

PR #525, opened 2022-04-06
maint: avoid regression/breaking change in `click` and declare our dependency to the library explicitly

consideRatio:
dask-gateway was dependent on click, but we didn't declare that dependency explicitly but instead relied on it showing up as a dependency from something else we had installed. Since we have a direct `import click` statement in the dask-gateway project, I suggest it is listed explicitly as a dependency like done in this PR now.

As part of listing `click` as an explicit dependency, we also pin it to avoid 8.1.0-8.1.2 which breaks things for us as discussed in #522.

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -7,6 +7,7 @@ npm install -g configurable-http-proxy
 # entries in ../dev-environment.yaml
 pip install -U \
     aiohttp \
+    "click<8.1.0" \
     colorlog \
     cryptography \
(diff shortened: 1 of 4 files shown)


consideRatio:
Aaarrgh, I have to pin click in four separate places!!! I think its absolutely critical that we de-duplicate the dependencies we list for maintenance to be sustainable.

I'll go for a self-merge on this so I can rebase #519 and avoid this added complexity to that PR.

PR #529, opened 2022-04-16
Add /api/v1/routes_initialized to help proxy process report readiness

consideRatio:
This resolves part of #422, an intermittent error stemming from `404 NOT FOUND` responses.

The issue was that the Golang proxy binary that was started in a separate process was only partially awaited. It was awaited to contact `/api/v1/routes` but not to have initialized its routes based on a response. See https://github.com/dask/dask-gateway/issues/422#issuecomment-1100606030 for investigation notes about this.

## This PR: await notification

In this PR I've opted to resolve this by a quite robust implementation where the Golang proxy binary running in a separate process reports that its routes are initialized after they are. Doing so comes with the complexity of adding an additional web-server API endpoint and making another web request from the Golang proxy binary.

```diff
     async def __aenter__(self):
         self.gateway = DaskGateway(config=self.config)
         self.gateway.initialize([])
         await self.gateway.setup()
-        await self.gateway.backend.proxy._proxy_contacted
+        await self.gateway.backend.proxy._proxy_routes_initialized
```

Closes #530

## An alternative PR: sleep some (#530)

An alternative to this implementation would be to sleep some duration, for example after the previous `_proxy_contacted` flag was awaited to be set. If we would go for this approach, maybe we test with a sleep of 0.25 seconds, then double it until we no longer observe these intermittent tests, and make a comment linking to this then closed PR about the motivation for sleeping.

```diff
     async def __aenter__(self):
         self.gateway = DaskGateway(config=self.config)
         self.gateway.initialize([])
         await self.gateway.setup()
         await self.gateway.backend.proxy._proxy_contacted
+        await asyncio.sleep(0.25)
```

### DIFF
--- a/dask-gateway-server/dask-gateway-proxy/main.go
+++ b/dask-gateway-server/dask-gateway-proxy/main.go
@@ -254,6 +254,32 @@ func (c *apiClient) getRoutingTable() (*RoutesMsg, error) {
 	return &msg, err
 }
 
+// reportRoutesInitialized helps us communicate to the dask-gateway-server that
+// we have acquired an initial set of routes and can from now on act based on
+// them without responding with 404 NOT FOUND.
+func (c *apiClient) reportRoutesInitialized() (*RoutesMsg, error) {
+	req, err := http.NewRequest("GET", c.apiURL+"_initialized", nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Authorization", "token "+c.apiToken)
+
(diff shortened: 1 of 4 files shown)


consideRatio:
@martindurant @jacobtomlinson agree to merge #530 over this more complicated thing that introduced changes that can cause more maintenance burden? In #530 we now reference this strategy in a comment instead next to the `sleep` call made there. If we find that we have trouble with 404 responses by using the sleep duration of 0.25, we double it and re-evaluate if we need to increase the sleep too much in time.

PR #530, opened 2022-04-16
Fix intermittent errors by sleeping a bit before running tests

consideRatio:
This is an alternative approach to #529 to avoiding the ValueError 404 NOT FOUND as described in #422.

Feel free to merge either alternative. Refer to #529 for an overview of the choice. Closes #529.

Note that with a sleep duration of 0.25, we have not experienced any failures related to 404 NOT FOUND in 3/3 workflows executions.


consideRatio:
I want to see this together with #536 fixes the PBS tests also that experienced memory issues.

PR #533, opened 2022-04-16
Helm chart images: conda removed -> pip only, usage disclaimer added, minimized Dockerfile complexity

### LINKED ISSUE #432: Update to use `mamba` instead of `conda` in Dockerfiles
### Update

No consensus to make a shift to use `mamba` instead of `conda` currently within these Dockerfiles.

### Original proposal

I suggest we update to use `mambaforge` to setup `conda` and such in our Dockerfiles. It helps speed up installation by being a conda replacement with higher performance.

There have been some quirks about a Python version updated that I'll be on the lookup in the unlikely case it would happen, otherwise it has been a great success to transition to use `mamba` in projects like docker-stacks and repo2docker with very significant performance improvements to build images.

### LINKED ISSUE #489: Make conda/mamba flavor a build arg
I know there was some discussion of using mamba in #432.  Due to the change in anaconda.org's licensing agreements, we prefer to use miniforge/mambaforge whenever possible.

A patch like this: https://github.com/holzman/dask-gateway/commit/96b714515a528836f71d4a2c58adf10649af9d71 seems like it should make everyone happy.

### LINKED ISSUE #524: Reduce complexity of Helm chart's Dockerfiles and define expectations
We have two Dockerfiles that we build and use within the Helm chart:

- [dask-gateway-server/Dockerfile](https://github.com/dask/dask-gateway/blob/main/dask-gateway-server/Dockerfile): used by the `controller` and `api` pods.
- [dask-gateway/Dockerfile](https://github.com/dask/dask-gateway/blob/main/dask-gateway/Dockerfile): used by default for `scheduler` and `worker` pods.

### Reduce complexity - respect setup.py's `install_requires`

I'd like to make these images install `dask-gateway` and `dask-gateway-server` in a more straight forward manner. Currently we first install `conda-forge` provided dependencies matching those we know we need by a hardcoded list and ignore the dependency details specified in setup.py.

This adds complexity to this project! For example in https://github.com/dask/dask-gateway/issues/522#issuecomment-1089478613 I thought that `setup.py` defined restrictions on a dependency that caused test failures for us would be respected also in the images we build for our k8s tests - but it wasn't and I kept looking for other reasons of failing k8s tests.

### Document expectations - how they intended to be used?

These images are around, but it isn't clear for us maintainers or end users how they are to be used. My perception is that end users are not to derive their own images from these, but instead install `dask-gateway` with a package manager in their own image whenever they want to customize a scheduler or worker image.
  
In practice, I think we should declare that the images are not meant to be used as base images, nor are the Dockerfile's, which instead describe that users are supposed to do a conda-forge installation in their custom image to acquire a `scheduler` and a `worker` suitable image.

### Related

- I've have tried to avoid duplicated specification of Python dependencies as the maintenance complexity that follows spreading them out is high
  #516
  #519
- I failed to realize that dependencies in `setup.py` wasn't respected in these images as `--no-deps` is specified, and a set of hardcoded conda packages are installed ahead of time as a separate set of dependencies.
  #522
- There have been discussion related to end users use of the repo's dask-gateway/Dockerfile
  #489
  
## Action points

- [x] Discuss: is it acceptable to lower the expectations and asking users to not rely on these images?
- [x] Create a PR related to https://github.com/dask/dask-gateway/issues/524#issuecomment-1100384055 including documentation updates, a disclaimer in the Dockerfiles, etc.
  #533

consideRatio:
Discussion for this change was made in https://github.com/dask/dask-gateway/issues/524#issuecomment-1100384055.

>1. dask/dask-gateway-server, an image used by the Helm chart's controller and api pod, should not be something end users derive their own image from. If they do for some unknown reason though, they shouldn't expect this project to respect backward compatibility or changelogs etc.
>2. dask/dask-gateway, an image used by the Helm chart as a sample image for a clusters scheduler/worker pods, should be treated as a sample image that people shouldn't derive their own scheduler/worker images from. Instead, they should build their scheduler/worker images in any way [according to documentation](https://gateway.dask.org/install-kube.html#using-a-custom-image) even though that documentation may require an update.
>3. The example Dockerfile, just an example and not an image we build, is probably a bit too complicated to maintain and keep updated. As outdated examples are sometimes problematic I think the best course of action is to rely on the documentation on how to create your own image, and suggest asking users to look at the dask/dask-gateway image for practical inspiration on how you may do it.

## Change summary

- Documents that the dask/dask-gateway image is meant for testing purposes and possibly inspiration of writing your own image. It also removes the example Dockerfile that had grown a bit outdated and tweaks documentation to reference the sample Dockerfile.
- Reduces the complexity of the Helm chart's Dockerfiles very significantly, cutting a CI job step down to around 68 and 80 seconds from around 284 and 316 seconds: a reduction of ~75% or ~3m46s. As the k8s tests were the by far the slowest in our test suite, this reduces our CI test duration with that amount.
- `conda` and `mamba` is no longer installed as part of these images, something I'm quite strongly opinionated about at the moment. If the Helm chart was a separate project that could build these images after dask-gateway and dask-gateway-server has been released and packaged to conda-forge, I'd think differently as the complexity comes mainly from that. I think this change is so problematic as there is now a more explicit expectation.
  - The dask/dask-gateway-server running in the `api` pod and `controller` pod isn't something users have overridden historically I think
  - The dask/dask-gateway image for scheduler/worker pods is documented as a sample image used for testing with an expectation that users define their own image.

Closes #524
Closes #489 - by declaring that the user is to make their own images for scheduler/workers if they have specific needs.
Closes #432 - by going for plain pip instead.


dask-gateway-server/Dockerfile
```diff
@@ -1,45 +1,41 @@
...
+#
+FROM python:3.10-slim-bullseye
+
+# Install tini and upgrade linux packages are updated to patch known
+# vulnerabilities.
 RUN apt-get update \
-    && apt-get install -y tini \
-    && rm -rf /var/lib/apt/lists/*
-
+ && apt-get upgrade -y \
```
martindurant, line 14: A full upgrade seems heavy handed - isn't equivalent to starting with a newer distribution?

dask-gateway-server/Dockerfile
```diff
@@ -1,45 +1,41 @@
...
-COPY --from=builder /tmp/install-prefix/lib /usr/local/lib/
-
-# Create non-root user and working directory
-WORKDIR /srv/dask-gateway
-RUN useradd -m -U -u 1000 dask && chown dask:dask /srv/dask-gateway
-USER 1000:1000
+        --requirement=/srv/dask-gateway-server/requirements.txt \
+        kubernetes-asyncio \
+ && pip install --no-cache-dir --no-deps \
+        /srv/dask-gateway-server --install-option="--no-build-proxy"
```
martindurant, line 38: Maybe delete pip's cache before completing this step

dask-gateway-server/requirements.txt
```diff
@@ -0,0 +1,9 @@
+# These are dependencies for installing the dask-gateway-server package.
+#
+# NOTE: changes to the dependencies here must also be reflected in
+#       ../dev-environment.yaml
+#
+aiohttp
```
martindurant, line 6: Version bounds?

dask-gateway/Dockerfile
```diff
@@ -1,95 +1,36 @@
-# ** A base miniconda image **
-FROM debian:bullseye-slim as miniconda
-LABEL MAINTAINER="Jim Crist-Harif"
-
-# List of conda versions: https://repo.anaconda.com/miniconda/
+# USAGE DISCLAIMER:
```
martindurant, line 1: Is this text a disclaimer?

dask-gateway/requirements.txt
```diff
@@ -0,0 +1,16 @@
...
+#       ../dev-environment.yaml
+#
+aiohttp
+# FIXME: click 8.0.4 works, but 8.1.0-8.1.2 has found to cause failures for
+#        currently unknown reasons.
+#
+#        This is tracked in https://github.com/dask/dask-gateway/issues/522.
+#
+click<8.1.0
+dask>=2.2.0
```
martindurant, line 13: Do we need to ensure that dask and distributed have matching versions?

docs/source/install-kube.rst
```diff
@@ -216,9 +216,9 @@ There are no other requirements for images, any image that meets the above
 should work fine. You may install any additional libraries or dependencies you
 require.
 
-To develop your own image, you may either base it on a compatible version of
-``daskgateway/dask-gateway``, or use our `example dockerfile`_ as a reference
-and develop your own.
+We encourage you to maintain your own image for scheduler and worker pods as
+this project only provides a `minimal image`_ for testing purposes which can be
+used as inspiration, but isn't recommended to be used as a base image.
```
martindurant, line 221:
```suggestion
used as inspiration, but isn't recommended to be used in production.
```

martindurant (commented):
Minor thoughts only. I wonder if you can compare the resultant pip-based image size with conda - is it a significant savings?

One reason for including conda (/mamba) is that users might expect it to be something they can call from within workers. Or maybe not.

dask-gateway-server/Dockerfile
consideRatio: Ah no this is quite trivial, it will only do small upgrades for known vulnerabilities in my experience from doing it in jupyterhub/zero-to-jupyterhub-k8s. The official images on dockerhub are slow to update, so this can be helpful for such situations.

consideRatio:
Thanks a lot for your review @martindurant!!!!

I'm headed away so I can't go through it fully now so I'll return to it tomorrow/later. Cheers!

docs/source/install-kube.rst
jcrist: It can be used in production, and I wouldn't even say we don't recommend using it as a base image, it's just that we won't do anything extra to support using it as a base image (adding extra build args, etc...).

dask-gateway/requirements.txt
jcrist: That shouldn't actually matter, but it should end up happening in practice (dask & distributed's version constraints + pip chosing the most recent releases).

jcrist (approved):
Just a few small nits, otherwise this looks good to me. Thanks Erik!

dask-gateway-server/requirements.txt
consideRatio: The versions pinning seen here are unchanged from being extracted from setup.py, do you suggest we put bounds like <4 here or similar?

I've understood it as a bad practice to constrain the versions of a python package without knowing ahead of time its a problem and are used to just declaring upper bounds when its know to cause issues that are to be fixed but aren't yet fixed.

martindurant: I imagine there are versions that just don't work (any more); but sometimes it's useful to guard against unknown future breakages. But since this was copied, perhaps this isn't the time. Also it's hard to do thorough research on compatible versions.

dask-gateway/Dockerfile
consideRatio: Hmmmm I'm not sure. Googling the definition of disclaimer led me to:

> a statement that denies something, especially responsibility.

This is the part that I was thinking of as a disclaimer, it is quite indirect, it is a suggestion that we don't want to be held responsible for meeting the needs of the users by asking them to build their own.

> Please build your own according to
# the documentation if this very limited image doesn't meet your needs.

---

I'm not confident on a change to make here but I'm very happy to accept any change suggestions to this text.

dask-gateway/requirements.txt
consideRatio: This file represents the dependencies of the dask-gateway package, I'm thinking as long as `dask-gateway` doesn't have requirements to match something for its own function we shouldn't constrain it here. I think it becomes up to the person defining the following environments to ensure things match.
- where a dask-gateway client runs
- where the scheduler runs
- where the worker runs

We provide a sample environment (dask-gateway Dockerfile) that represent the scheduler and worker, but doesn't manage an environment where the dask-gateway client runs.

I guess the dask-gateway client should be compatible with the dask-gateway-server as well if there is a breaking change making client/server incompatible with each other.

I presume there is some documentation in the dask ecosystem about the relevance of aligning versions to some degree for certain packages etc? Is it perhaps so that the only additional complexity introduced by dask-gateway is the possebility of a need to align dask-gateway the client with dask-gateway-server?

docs/source/install-kube.rst
consideRatio, line 221:
```suggestion
We encourage you to maintain your own image for scheduler and worker pods as
this project only provides a `minimal image`_ for testing purposes.
```

I made this suggestion based on the discussion just above in https://github.com/dask/dask-gateway/pull/533/files#r852284873.

dask-gateway-server/requirements.txt
consideRatio: Ah, yes lower bounds would be good to maintain. I think it's a very relevant improvement to aim for. I opened #539 about this!

dask-gateway-server/Dockerfile
consideRatio: I don't think there are any persistent cache to clear thanks to `--no-cache-dir`. When I build this image with `--no-cache-dir` I see its showing logs like...

```
Created wheel for tornado: filename=tornado-6.1-cp310-cp310-linux_x86_64.whl size=414482 sha256=1a5449b7d8c257cff2f744b63710c3bea30786d0e6ba16d1a1600ec9c915cc38
  Stored in directory: /tmp/pip-ephem-wheel-cache-hzbv0mzj/wheels/80/32/8d/21cf0fa6ee4e083f6530e5b83dfdfa9489a3890d320803f4c7
```

Inspecting the image from a started container I don't see that folder around either.

```shell
# my computers terminal
docker run -it --rm --entrypoint bash test

# the started containers terminal inspected
dask@7d8fe549ecf5:~$ pwd 
/home/dask
dask@7d8fe549ecf5:~$ ls -alh .
total 24K
drwxr-xr-x 1 dask dask 4.0K Apr 18 21:55 .
drwxr-xr-x 1 root root 4.0K Apr 16 15:39 ..
-rw-r--r-- 1 dask dask  220 Aug  4  2021 .bash_logout
-rw-r--r-- 1 dask dask 3.5K Aug  4  2021 .bashrc
drwxr-xr-x 4 dask dask 4.0K Apr 18 21:55 .local
-rw-r--r-- 1 dask dask  807 Aug  4  2

dask@7d8fe549ecf5:~$ ls -alh /tmp
total 268K
drwxrwxrwt 1 root root 4.0K Apr 18 21:55 .
drwxr-xr-x 1 root root 4.0K Apr 18 21:56 ..
-rw------- 1 root root 260K Apr 14 04:29 tmpvj26enrfcacert.pem
```

dask-gateway-server/Dockerfile
consideRatio: This is what changed when I did built this image:

```Dockerfile
FROM python:3.10-slim-bullseye

RUN apt-get update \
 && apt-get upgrade -y
```

```
Step 1/2 : FROM python:3.10-slim-bullseye
 ---> ea290afaeb03
Step 2/2 : RUN apt-get update  && apt-get upgrade -y
 ---> Running in 99bfddeb5dc1
Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]
Get:2 http://security.debian.org/debian-security bullseye-security InRelease [44.1 kB]
Get:3 http://deb.debian.org/debian bullseye-updates InRelease [39.4 kB]
Get:4 http://deb.debian.org/debian bullseye/main amd64 Packages [8182 kB]
Get:5 http://security.debian.org/debian-security bullseye-security/main amd64 Packages [126 kB]
Get:6 http://deb.debian.org/debian bullseye-updates/main amd64 Packages [2596 B]
Fetched 8510 kB in 2s (3930 kB/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
Calculating upgrade...
The following packages will be upgraded:
  gzip liblzma5
2 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Need to get 300 kB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 http://security.debian.org/debian-security bullseye-security/main amd64 gzip amd64 1.10-4+deb11u1 [132 kB]
Get:2 http://security.debian.org/debian-security bullseye-security/main amd64 liblzma5 amd64 5.2.5-2.1~deb11u1 [168 kB]
debconf: delaying package configuration, since apt-utils is not installed
Fetched 300 kB in 1s (469 kB/s)
(Reading database ... 7023 files and directories currently installed.)
Preparing to unpack .../gzip_1.10-4+deb11u1_amd64.deb ...
Unpacking gzip (1.10-4+deb11u1) over (1.10-4) ...
Setting up gzip (1.10-4+deb11u1) ...
(Reading database ... 7023 files and directories currently installed.)
Preparing to unpack .../liblzma5_5.2.5-2.1~deb11u1_amd64.deb ...
Unpacking liblzma5:amd64 (5.2.5-2.1~deb11u1) over (5.2.5-2) ...
Setting up liblzma5:amd64 (5.2.5-2.1~deb11u1) ...
Processing triggers for libc-bin (2.31-13+deb11u3) ...
Removing intermediate container 99bfddeb5dc1
```

consideRatio:
@martindurant @jcrist thanks for your review! I've responded to the review comments! I know that this and #538 will have merge conflicts so I'll need to do some work as soon as one is merged and I'm quite anxious to keep moving.

consideRatio:
> Minor thoughts only. I wonder if you can compare the resultant pip-based image size with conda - is it a significant savings?

No savings in size on the image that I think matters to compare, dask-gateway-server. The build time of both images are significantly reduced though.

```
dg                                            latest               d0b41e6c4250   2 minutes ago        548MB
dg-new                                        latest               a0e4c7c492a8   28 seconds ago       163MB

dgs                                           latest               636d4a82d66b   About a minute ago   189MB
dgs-new                                       latest               c5a1bd202a43   4 seconds ago        193MB
```

martindurant:
Big win on one, and slightly worse on the other. Interesting! The big win is a good argument to go ahead.

consideRatio:
Hey @martindurant, thanks for your attention!

This PR blocks work to fix merge conflicts in #538 and would help having merged to continue development in #514 where I have a not yet understood test failure and I hope to rule out it's related to the old Dockerfiles. Please ping me for any followup questions or and I'll work to address them with haste!

dask-gateway/Dockerfile
martindurant:
```suggestion
# USAGE:
```

(or perhaps "purpose", "description". To me "disclaimer" sounds like "warning".

dask-gateway/requirements.txt
martindurant: I suppose having the REST endpoints like /v1/ is a reasonable defence against breaking changes.

docs/source/install-kube.rst
martindurant: ok

martindurant (approved):
All good, happy with those conversations. Feel free to pick your favourite wording around the "production" image

consideRatio:
Thank you @martindurant!! I updated `USAGE DISCLAIMER:` to be `PURPOSE:` instead

martindurant:
👍

PR #538, opened 2022-04-18
ci, maint: build/publish to PyPI, build linux/mac and amd64/arm64 wheels

consideRatio:
## Summary

- We now have automation to publish packaged source distributions and wheels to PyPI.
- We provide four platform dependent wheels for dask-gateway-server for linux and mac that comes in amd64 and arm64 variants.
- Various refactoring is done and documentation is improved to reduce complexity and improve long term maintenance.
- This workflow was developed against PyPI's test server, see https://test.pypi.org/project/dask-gateway-server/0.9.0/#files for the result.

## Noteworthy details

- [dask-gateway-server/build_tools](https://github.com/dask/dask-gateway/tree/main/dask-gateway-server/build_tools) are removed as most relevant logic moved to dask-gateway-server/setup.py and the remaining logic to set `GOOS` and `GOARCH` Golang cross compilation environment variables made sense to do in the GitHub Workflow without having a dedicated script for.
- dask-gateway-server's `--no-build-proxy` flag in setup.py is replaced with an environment variable `DASK_GATEWAY_SERVER__NO_PROXY`. Retaining the flag forced many not-robust workarounds. The environment variable also helps us ensure we don't package a binary in wheels we build.
  
Closes #537


consideRatio:
I'll go for a merge and trial a release to verify functionality of the CI system and if it succeeds, follow up with tests against the release end-to-end in a real world installation.

PR #541, opened 2022-04-18
ci: avoid running tests in fork's PR branches

consideRatio:
If tests are run in a fork, they will fail, as they don't have access to the semi-private testing images. When that fail happen, it will be seen as a red cross on the commit even before the pull_request triggered tests has completed. Like this, we save some compute by not running things twice and we avoid test failure notifications etc.

### About semi-private images

The `ghcr.io/dask/dask-gateway-ci-hadoop` image and other `ghcr.io/dask/dask-gateway-ci-` images are marked as private, but we have also made them available to PRs in this repo. So, they are private to not draw attention rather than being secret.

### Screenshots from errors we avoid with this PR

image

image

image

### DIFF
--- a/.github/workflows/build-publish-docs.yaml
+++ b/.github/workflows/build-publish-docs.yaml
@@ -16,9 +16,7 @@ on:
       - "dask-gateway/**"
       - "dask-gateway-server/**"
       - ".github/workflows/build-publish-docs.yaml"
-    branches-ignore:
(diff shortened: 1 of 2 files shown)


jcrist (approved):
LGTM, thanks Erik!

consideRatio:
Thanks Jim!

PR #554, opened 2022-04-21
Update traefik's CRDs

consideRatio:
To update Traefik's CRDs is a breaking change that requires manual intervention from users with an existing installation of the dask-gateway helm chart, due to that, I want to update them to the latest version before release.

### DIFF
--- a/resources/helm/dask-gateway/crds/daskclusters.yaml
+++ b/resources/helm/dask-gateway/crds/daskclusters.yaml
@@ -1,3 +1,7 @@
+# Any change to these resources should be followed by an entry in the changelog
+# about needing to manually patch them, read more about this in
+# https://github.com/dask/dask-gateway/issues/553.
+#
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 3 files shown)


dhirschfeld:
I'll just mention, in case you haven't already seen, the CRDs can be updated with e.g.
```
kubectl apply --server-side --force-conflicts -k https://github.com/traefik/traefik-helm-chart/tree/v19.0.4/traefik/crds
```
***xref:*** https://github.com/traefik/traefik-helm-chart/pull/649

Copying locally might still be easier for your users though.

consideRatio:
Thanks @dhirschfeld!

consideRatio:
@dhirschfeld do you grasp if `--server-side --force-conflicts` makes sense to suggest overall when doing `kubectl apply` of the bundled CRDs?

dhirschfeld:
TBH, I'm no expert here and I don't really have a good grasp on the differences between a server-side and a normal apply.

My understanding of `--force-conflicts` is that it will allow making changes to existing fields which we'd definitely want to do here. I'm not too sure when you'd want an apply to *not* make changes - maybe if an apply was supposed to only add new fields and not update any existing ones having the update error might be useful.

This doc has a pretty good/understandable explanation:
* https://kubernetes.io/docs/reference/using-api/server-side-apply/

...and doing some more research brought up this blog which has a bit more context:
* https://opensource.googleblog.com/2021/10/server-side-apply-in-kubernetes.html

It sounds like it's supposed to be a new, better (show diffs, detect conflicts, ???) way to do a `kubectl apply`

consideRatio:
Thanks for helping out with the learning process, I'm leaning towards thinking we should suggest both `--server-side` and `--force-conflicts`

dhirschfeld:
> *I'm leaning towards thinking we should suggest both `--server-side` and `--force-conflicts`*

I *think* that's the way to go also

PR #559, opened 2022-04-21
golang: refresh dask-gateway-proxy using modules and package directories

### LINKED ISSUE #506: Refresh dask-gateway-server/dask-gateway-proxy with modern golang practices
With the current versions of golang (1.17, 1.18), golang has established some practices on how to handle dependencies etc that wasn't around in the past. To run the tests against golang 1.16-1.18 I had to make a small change in #505 for example.

I'd like for the golang code we maintain as part of this project to catch up and follow common golang code practices to help it be easy to comprehend in the context of modern golang documentation and practices etc. Due to that, I suggest we try to refactor the source code in dask-gateway-server/dask-gateway-proxy to follow a common practice for how the project is structured: https://github.com/golang-standards/project-layout#go-directories

I'd be happy to do this, but I need to learn more about Golang first, which will take quite a bit of time. I think if anyone wants to contribute a PR to refactor our golang source code to follow common practices such as how to organize files etc, that would be great!

rigzba21:
@consideRatio Restructuring dask-gateway-proxy using [go project layout](https://github.com/golang-standards/project-layout#go-directories) to address https://github.com/dask/dask-gateway/issues/506

Nothing has changed as far as functionality. 
- Both `router.go` and `sni.go` have been moved into separate packages under the `pkg/` directory. 
- `main.go` has been moved to `cmd/dask-gateway-proxy`. 
- `logging.go` has been moved to its own internal package under `internal/logging`.  
- Generated a new `go.mod` and `go.sum` with dependencies.
- GitHub Actions updated to point to `pkg/`, `cmd/dask-gateway-proxy/`, and `internal/`. 
  - _Currently there are only unit tests for `pkg/router`._
  
I hope this helps!

---

EDIT: closes #506!


dask-gateway-server/dask-gateway-proxy/pkg/sni/sni.go
```diff
@@ -31,7 +31,7 @@ func (c *peekedTCPConn) Read(p []byte) (n int, err error) {
 	return c.TCPConn.Read(p)
 }
 
-func wrapPeeked(inConn *net.TCPConn, br *bufio.Reader) tcpConn {
+func wrapPeeked(inConn *net.TCPConn, br *bufio.Reader) TcpConn {
```
consideRatio, line 34: Is it correct that you adopt MixedCase, a combination of camelCase and PascalCase, where...

> If an identifier needs to be visible outside the package, its first character should be uppercase.

?

dask-gateway-server/dask-gateway-proxy/cmd/dask-gateway-proxy/main.go
```diff
@@ -541,6 +545,7 @@ func (p *Proxy) runTCP(ln net.Listener, forwarder *Forwarder) {
 	}
 }
 
+//TODO: make this a skinny main?
```
consideRatio, line 548: Could you reword this todo a bit so I can understand what you mean? Now I understand it only as "refactor to include less logic", but I'm not sure on any detail or the benefits are. I would agree there are benefits to doing so if that is a common practice, because that makes the code easier to navigate for experienced golang developers for example.

dask-gateway-server/dask-gateway-proxy/tools/fail_without_go_1_12.go
```diff
@@ -1,3 +1,4 @@
+//go:build !go1.12
 // +build !go1.12
```
consideRatio, line 2: It would be perfectly fine to require golang 1.17+ for example. I think overall, declaring that we require 1.12 but not testing against 1.12 can just be a reason for possible issues.

Currently I know we test compile the binary with 1.16, 1.17, and 1.18, and know from https://endoflife.date/go that only 1.17+ hasn't reached end of life.

https://github.com/dask/dask-gateway/blob/01eef766f05d3ff91718b134ccb9490b02f50671/.github/workflows/test.yaml#L58-L63

consideRatio (approved):
Thank you for your work on this @rigzba21!!! From what I can tell, this looks good. I left a few comments for my own insight but this LGTM from what I can tell.

Also note that the test failures are unrelated at least to this PR (https://github.com/ipython/traitlets/issues/731).

dask-gateway-server/dask-gateway-proxy/cmd/dask-gateway-proxy/main.go
rigzba21: You are correct that it is a common pattern of including less logic in the main function, though I think I'll get rid of the comment. With the scope of this PR and the current size of the dask-gateway-proxy go code, it wouldn't make it more beneficial to refactor `main` at this time.

dask-gateway-server/dask-gateway-proxy/pkg/sni/sni.go
rigzba21: Correct. See [go spec docs on Exported Identifiers](https://go.dev/ref/spec#Exported_identifiers) for reference. 
>An identifier may be exported to permit access to it from another package. An identifier is exported if both: 
>1. the first character of the identifier's name is a Unicode upper case letter (Unicode class "Lu"); and

>2. the identifier is declared in the [package block](https://go.dev/ref/spec#Blocks) or it is a [field name](https://go.dev/ref/spec#Struct_types) or [method name](https://go.dev/ref/spec#MethodName).

consideRatio: Thanks for helping me learn about this @rigzba21!

dask-gateway-server/dask-gateway-proxy/tools/fail_without_go_1_12.go
rigzba21: I currently don't have a good enough understanding of the purpose of `fail_without_go_1_12.go` or where/how it's intended to be used, which is why I relocated it to the `tools/` directory.

consideRatio: Ah, neither do i. Perhaps it was a known requirement to have 1.12 at some point, and meant to provide a helpful failure message?

Good to know you dont recognize this practice, i figure then it makes most sense to just remove it at this point then.

consideRatio: Doing that is feature creep, this is good to go as it is i think as well.

consideRatio:
@martindurant @TomAugspurger @jacobtomlinson a ping for review, i figure we merge this soonish without further comments.

TomAugspurger:
I've never written go before, but +1 to the goals of this PR.

jacobtomlinson (commented):
This seems like a great move to me. The only comment I have is why are we putting some code under `pkg`? I don't think we should be intentionally exposing any public APIs and we should probably place it in `internal` instead.

rigzba21:
> The only comment I have is why are we putting some code under `pkg`? I don't think we should be intentionally exposing any public APIs and we should probably place it in `internal` instead.

I thought about that, however, I figured that might be a design decision outside of the scope of this PR, and might fit better in a separate PR.

consideRatio:
> I figured that might be a design decision outside of the scope of this PR, and might fit better in a separate PR

Deal! Let's go for a merge on this. The tests were passing before we ran into an issue I think was related to traitlets (https://github.com/ipython/traitlets/issues/731).

consideRatio:
@rigzba21 thank you so much for your work on this!!!!!! :tada: :heart: :sunflower:

PR #564, opened 2022-05-11
Document setting `display: False` in hub services config

yuvipanda:
Otherwise, a very confusing dask-gateway entry is provided under
'services' in the hub home page for the user.

Ref https://github.com/2i2c-org/infrastructure/issues/1301


yuvipanda:
@jcrist indeed, it wasn't documented in the JupyterHub docs! I just added that in https://github.com/jupyterhub/jupyterhub/pull/3895

choldgraf:
@jcrist none of us knew it existed 😅

Issue #568, opened 2022-06-05
Dashboard is unavailable for namespaces different from the gateway


olivier-lacroix:
Hello,

Thanks for making this great chart available!

When creating a dask cluster in a namespace different from the one the gateway itself is in, accessing the dask dashboard yields a 404 error. It works as expected when in the same namespace.

The following is logged by traefik

```
level=error msg=""Cannot create service: subset not found"" serviceName=dask-xxx providerName=kubernetescrd ingress=dask-xxx servicePort=""{0 8786 }"" namespace=custom-dask-namespace"
level=error msg=""Failed to create middleware keys: middleware dask-gateway/clusters-prefix-dask-gateway is not in the IngressRoute namespace custom-dask-namespace"" providerName=kubernetescrd ingress=dask-xxx namespace=custom-dask-namespace"
level=error msg=""the service \""custom-dask-namespace-dask-xxx-yyy@kubernetescrd\"" does not exist"" routerName=custom-dask-namespace-dask-xxx-yyy@kubernetescrd entryPointName=web"
level=error msg=""Cannot create service: subset not found"" servicePort=""{0 8786 }"" namespace=custom-dask-namespace providerName=kubernetescrd ingress=dask-xxx serviceName=dask-xxx"
level=error msg=""Failed to create middleware keys: middleware dask-gateway/clusters-prefix-dask-gateway is not in the IngressRoute namespace custom-dask-namespace"" namespace=custom-dask-namespace providerName=kubernetescrd ingress=dask-xxx"
level=error msg=""Failed to create middleware keys: middleware dask-gateway/clusters-prefix-dask-gateway is not in the IngressRoute namespace custom-dask-namespace"" namespace=custom-dask-namespace providerName=kubernetescrd ingress=dask-xxx"
level=error msg=""Failed to create middleware keys: middleware dask-gateway/clusters-prefix-dask-gateway is not in the IngressRoute namespace custom-dask-namespace"" ingress=dask-xxx namespace=custom-dask-namespace providerName=kubernetescrd"
level=error msg=""Failed to create middleware keys: middleware dask-gateway/clusters-prefix-dask-gateway is not in the IngressRoute namespace custom-dask-namespace"" providerName=kubernetescrd ingress=dask-xxx namespace=custom-dask-namespace"
[...]

PR #569, opened 2022-06-06
Let Traefik's route traffic across namespaces via IngressRoute resources

olivier-lacroix:
Hello,

Setting allowCrossNamespace=true fixes https://github.com/dask/dask-gateway/issues/568

see https://doc.traefik.io/traefik/providers/kubernetes-crd/#allowcrossnamespace

Not 100% sure of the security implication of that setting - would it be preferable to gate it behind a chart parameter?

Cheers,

Olivier

### DIFF
--- a/resources/helm/dask-gateway/templates/traefik/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/traefik/deployment.yaml
@@ -44,6 +44,7 @@ spec:
             - "--global.sendanonymoususage=False"
             - "--ping=true"
             - "--providers.kubernetescrd"
+            - "--providers.kubernetescrd.allowCrossNamespace=true"
             - '--providers.kubernetescrd.labelselector=gateway.dask.org/instance={{ include "dask-gateway.fullname" . }}'
             - "--providers.kubernetescrd.throttleduration=2"
             - "--log.level={{ .Values.traefik.loglevel }}"


consideRatio:
Ah nice investigative work! Is this tested to work already?

olivier-lacroix:
@consideRatio , yes, pulling the chart and patching it with that additional argument solves it for me.

olivier-lacroix:
Indeed, the tests don't look too good!

consideRatio:
@olivier-lacroix excellent okay!

Since users don't create IngressRoute objects themselves, but dask-gateway does it for them to specifically send traffic to the users created dask-cluster resources (scheduler/workers), I think we should be fine. If we didn't allow cross-namespaced work, I'd argue we should lower the permissions granted to traefik as well, but we have instead granted significant permissions.

https://github.com/consideRatio/dask-gateway/blob/7d16965b300d45d1ae0f39ce816818070a24545a/resources/helm/dask-gateway/templates/traefik/rbac.yaml#L8-L19

olivier-lacroix:
> @olivier-lacroix excellent okay!
> 
> Since users don't create IngressRoute objects themselves, but dask-gateway does it for them to specifically send traffic to the users created dask-cluster resources (scheduler/workers), I think we should be fine. If we didn't allow cross-namespaced work, I'd argue we should lower the permissions granted to traefik as well, but we have instead granted significant permissions.
> 
> https://github.com/dask/dask-gateway/blob/01eef766f05d3ff91718b134ccb9490b02f50671/resources/helm/dask-gateway/templates/traefik/rbac.yaml#L8-L19
> 
> I think the intention was to support this by default historically, but with modern versions of traefik, allowing this by default changed in v2.3.5: [traefik/traefik#7595](https://github.com/traefik/traefik/pull/7595)

OK, great then!

resources/helm/dask-gateway/templates/traefik/deployment.yaml
```diff
@@ -44,6 +44,7 @@ spec:
             - "--global.sendanonymoususage=False"
             - "--ping=true"
             - "--providers.kubernetescrd"
+            - "--providers.kubernetescrd.allowCrossNamespace=true"
             - '--providers.kubernetescrd.labelselector=gateway.dask.org/instance={{ include "dask-gateway.fullname" . }}'
             - "--providers.kubernetescrd.throttleduration=2"
```
consideRatio, line 49: Hmmm, one confusion point for me is case sensitivity.

Is this accepted? Looking at https://doc.traefik.io/traefik/providers/kubernetes-crd/#allowcrossnamespace, it could make more sense if this was `--providers.kubernetesCRD.allowCrossNamespace=true` or `--providers.kubernetescrd.allowcrossnamespace=true` in a way.

I wonder if there is a silent issue with this or other flags? I saw though in the logs of traefik that this was enabled, but with a warning asking if it was intentional or not, which seems weird if we explicitly try to set it.

I'm confused about how we pass these flags overall, hmmm...

consideRatio: I think everything is case insensitive, nevermind.

Issue #539, opened 2022-04-18
Investigate and set lower bounds of requirements for `dask-gateway` and `dask-gateway-server`


consideRatio:
In https://github.com/dask/dask-gateway/pull/533/files#r852422223 @martindurant noted the value of lower bounds for the requirements to install `dask-gateway` and `dask-gateway-server`.

I think that is a good practice and sustainable to maintain if it is tested in the CI system. When lower bounds are established, one can test them in the automatically like this [in jupyterhub](https://github.com/jupyterhub/jupyterhub/blob/c07358a52653dd0f3bccface7a11c6d36696b90f/.github/workflows/test.yml#L132-L143).

```yaml
      - name: Install Python dependencies
        run: |
          pip install --upgrade pip
          pip install --upgrade . -r dev-requirements.txt

          if [ "${{ matrix.oldest_dependencies }}" != "" ]; then
            # take any dependencies in requirements.txt such as tornado>=5.0
            # and transform them to tornado==5.0 so we can run tests with
            # the earliest-supported versions
            cat requirements.txt | grep '>=' | sed -e 's@>=@==@g' > oldest-requirements.txt
            pip install -r oldest-requirements.txt
          fi
```

PR #573, opened 2022-06-10
Fix compatibility with distributed >= 2022.5.1 and traitlets >= 5.2.0, and raise the lower bound of required versions

### LINKED ISSUE #571: No changes in this repo, but our tests are now failing
We have test failures and I'm not sure what they relate to, but maybe https://github.com/ipython/traitlets/issues/731. I've verified it was not introduced as part of a change in a merged PR, but because of a change over time. I know that May 24 they were broken, and may 9 they were functional. See https://github.com/dask/dask-gateway/actions/workflows/test.yaml?query=branch%3Amain

consideRatio:
Closes #571.

I've observed a flurry of issues that created a very hard to debug situation, but this resolves them all I think. This patch makes us require modern versions of `distributed`, `traitlets`, `click`, and `dask` though. Unless we have tests against the minimum required versions, I think its better that we do this than hope a version could be slightly lower without issues though.

---

The `click` issue as seen on a scheduler pod starting and quickly terminating.

```
Traceback (most recent call last):
  File "/srv/conda/envs/notebook/bin/dask-scheduler", line 8, in <module>
    sys.exit(go())
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/distributed/cli/dask_scheduler.py", line 216, in go
    check_python_3()
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/distributed/cli/utils.py", line 39, in check_python_3
    from click import _unicodefun
ImportError: cannot import name '_unicodefun' from 'click' (/srv/conda/envs/notebook/lib/python3.9/site-packages/click/__init__.py)
```

### DIFF
--- a/dask-gateway/requirements.txt
+++ b/dask-gateway/requirements.txt
@@ -3,6 +3,6 @@
 aiohttp
 click
 dask>=2.2.0
-distributed>=2021.01.1
+distributed>=2022.5.2
 pyyaml
 tornado
--- a/dask-gateway/requirements.txt
+++ b/dask-gateway/requirements.txt
@@ -1,8 +1,8 @@
 # These are dependencies for installing the dask-gateway package.
 #
 aiohttp
-click
-dask>=2.2.0
+click>=8.1.3
+dask>=2022.5.2
 distributed>=2022.5.2
 pyyaml
 tornado
--- a/dask-gateway/requirements.txt
+++ b/dask-gateway/requirements.txt
@@ -2,7 +2,7 @@
 #
 aiohttp
 click>=8.1.3
-dask>=2022.5.2
-distributed>=2022.5.2
+dask>=2022.1.0
+distributed>=2022.1.0,!=2022.5.1,!=2022.5.2
 pyyaml
 tornado
--- a/dask-gateway-server/dask_gateway_server/app.py
+++ b/dask-gateway-server/dask_gateway_server/app.py
@@ -13,7 +13,7 @@ from .auth import Authenticator
 from .backends import Backend
 from .routes import default_routes
 from .traitlets import Application, Type
-from .utils import AccessLogger, LogFormatter, classname, normalize_address, run_main
+from .utils import AccessLogger, classname, normalize_address, run_main
 
 
 class GenerateConfig(Application):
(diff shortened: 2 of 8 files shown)


consideRatio:
When we don't have distributed pinned below 2022.5.0, we get:

image

When we don't have traitlets pinned below 5.2.0, we get:

image

consideRatio:
It seems that in distributed 2022.05.1, https://github.com/dask/distributed/pull/6363 was introduced, and that removes a `fast` parameter from `scheduler.close()`. So, dask-gateway must use <=2022.05.0 and we should require >=2022.05.1 after we adjust to this removal.

consideRatio:
### Traitlets 5.2.0



### Traitlets 5.2.1

image

### Traitlets 5.2.2.post.1

TomAugspurger:
Haven't looked closely here, but linking to https://github.com/dask/distributed/pull/6561. With that PR to distributed, we can perhaps avoid changing the dask / distributed requirements. The current release will still be incompatible `distributed=2022.5.2`, but the current release will be compatible with older and newer versions (with these changes, it'll be compatible with those versions plus `distributed=2022.5.2`)

consideRatio:
It adds maintenance burden to ensure that things work with old versions, so keeping an old lower bound that we don't test doesn't feel great.

What should be done ideally, is to run tests against the oldest versions declared support for, like done in jupyterhub: https://github.com/jupyterhub/jupyterhub/blob/f6eec29aa222276ea8a9d8066eabb53e9f65b1a7/.github/workflows/test.yml#L132-L159.

dask-gateway/requirements.txt
```diff
@@ -1,8 +1,8 @@
 # These are dependencies for installing the dask-gateway package.
 #
 aiohttp
-click
-dask>=2.2.0
-distributed>=2021.01.1
+click>=8.1.3
+dask>=2022.5.2
+distributed>=2022.5.2
```
consideRatio, line 6:
```suggestion
dask>=2022.1.0
distributed>=2022.1.0,!=2022.5.1,!=2022.5.2
```

consideRatio: Btw, this discussion relates to https://github.com/dask/dask-gateway/issues/539.

consideRatio:
@TomAugspurger I'd like to go for a release and get this resolved today if possible, is https://github.com/dask/dask-gateway/pull/573#discussion_r895001185 reasonable to you?

TomAugspurger (approved):
Looks good thanks.

PR #583, opened 2022-06-29
Helm chart: add imagePullSecrets for gateway.backend

chadsr:
This PR adds `imagePullSecrets` support for the `gateway.backend` (`dask-gateway`) image used by schedulers and workers.

### DIFF
--- a/resources/helm/dask-gateway/templates/gateway/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/configmap.yaml
@@ -52,6 +52,8 @@ data:
             "%s:%s" % (image_name, image_tag) if image_tag else image_name
         )
 
+    c.KubeClusterConfig.imagePullSecrets = "{{ .Values.gateway.backend.imagePullSecrets }}"
+
     # Forward dask cluster configuration
     for field, prop_name in [
         # Scheduler config
--- a/resources/helm/dask-gateway/templates/gateway/configmap.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/configmap.yaml
@@ -52,7 +52,7 @@ data:
             "%s:%s" % (image_name, image_tag) if image_tag else image_name
         )
 
-    c.KubeClusterConfig.imagePullSecrets = "{{ .Values.gateway.backend.imagePullSecrets }}"
+    c.KubeClusterConfig.image_pull_secrets = {{ .Values.gateway.backend.imagePullSecrets }}
 
     # Forward dask cluster configuration
     for field, prop_name in [
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 
 from kubernetes_asyncio import client, config
 from kubernetes_asyncio.client.rest import ApiException
-from traitlets import Dict, Float, Unicode, default
+from traitlets import Dict, Float, Unicode, List, default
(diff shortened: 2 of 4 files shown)


resources/helm/dask-gateway/templates/gateway/configmap.yaml
```diff
@@ -52,6 +52,8 @@ data:
             "%s:%s" % (image_name, image_tag) if image_tag else image_name
         )
 
+    c.KubeClusterConfig.imagePullSecrets = "{{ .Values.gateway.backend.imagePullSecrets }}"
```
chadsr, line 55: Is this the correct way to go about mapping the property?
I saw the other `backend` values were mapped here, so I went for it.

consideRatio: - Look into if you see `KubeClusterConfig` which represents a Python class, have the configuration for `imagePullSecrets` then the left part of the assigment makes sense
- Note that `"{{ .Values.gateway.backend.imagePullSecrets }}"`` on the right side of `=` includes quotes, so the result will be `"[]"` which won't make sense - it should be a list, not a string.

consideRatio: This could be made to make sense, but you need to define `image_pull_secrets` as well it seems, like image_pull_policy is defined already.

```
c.KubeClusterConfig.image_pull_secrets = {{ .Values.gateway.backend.imagePullSecrets }}
```

https://github.com/dask/dask-gateway/blob/e6ee014730991730ede59cbeb6ebaded27b0d22f/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py#L51-L55

consideRatio:
## Support for the idea of this PR

Since we provide a dedicated option for the image pull policy and image name/tag etc for `gateway.backend`, I support adding `imagePullSecrets` next to it as well. But, we should not add everything we can configure as a dedicated configuration, then we should use the workaround below instead.

## A workaround

imagePullSecrets can be configured via the following config options:

https://github.com/dask/dask-gateway/blob/e6ee014730991730ede59cbeb6ebaded27b0d22f/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py#L143-L180

that are exposed via the Helm chart's `gateway.backend.scheduler.extraContainerConfig` and `gateway.backend.worker.extraContainerConfig` that you can pass `imagePullSecrets` to.

Since `gateway.backend.imagePullSecrets`

szachovy:
Hi,

Thanks @consideRatio for your input. I believe there should be `gateway.backend.scheduler.extraPodConfig` instead of `gateway.backend.scheduler.extraContainerConfig`

The minimal working proof:
```
jupyterhub:
  proxy:
    secretToken: "<token-1>"
  hub:
    services:
      dask-gateway:
        apiToken: "<token-2>"
dask-gateway:
  gateway:
    auth:
      jupyterhub:
        apiToken: "<token-2>"
    extraConfig:
      optionHandler: |
        from dask_gateway_server.options import Options, String
        def option_handler(options):
            return {
                "image": options.image,
                "scheduler_extra_pod_config": {"imagePullSecrets": [{"name": options.extra_pod_config}]},
            }
        c.Backend.cluster_options = Options(
            String("image", label="Image"),
            String("extra_pod_config", label="Image Pull Secrets"),
            handler=option_handler,
        )

```

If the scheduler and workers use the same image within the cluster, I am unsure whether to add `"worker_extra_pod_config": {"imagePullSecrets": [{"name": options.extra_pod_config}]}` to the returned function.

Best regards,
Wiktor

PR #590, opened 2022-07-29
Reorder gateway config and update comment about gateway.nodeSelector

GeorgianaElena:
This PR:

- relocates a few config options to have them just above the `gateway.backend` entry, otherwise the big `backend` config description makes it hard to see that they belong to gateway config and are not part of `gateway.backend`.
- updates the comment about what are the "gateway pods" that the `gateway.nodeSelector` will apply too.
  This actually caused me a bit of confusion, because I thought the `"gateway pods"` were actually the backend worker and scheduler pods, but the selector actually got applied to the `api` pod too ([reference](https://github.com/2i2c-org/infrastructure/pull/1573))

### DIFF
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -78,6 +78,20 @@ gateway:
     periodSeconds: 10
     failureThreshold: 3
 
+  # Settings for nodeSelector, affinity, and tolerations for the gateway pods,
+  # i.e., the `api` pod that runs the `dask-gateway-server`, and the
+  # backend worker/scheduler pods.
+  nodeSelector: {}
+  affinity: {}
+  tolerations: []
+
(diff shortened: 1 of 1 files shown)


resources/helm/dask-gateway/values.yaml
```diff
@@ -78,6 +78,20 @@ gateway:
     periodSeconds: 10
     failureThreshold: 3
 
+  # Settings for nodeSelector, affinity, and tolerations for the gateway pods,
+  # i.e., the `api` pod that runs the `dask-gateway-server`, and the
+  # backend worker/scheduler pods.
```
consideRatio, line 83:
```suggestion
  # nodeSelector, affinity, and tolerations the for the `api` pod running dask-gateway-server
```

consideRatio (approved):
Thanks @GeorgianaElena! This has confused me as well!!

PR #593, opened 2022-08-11
fix: add missing fields nameOverride and fullnameOverride

maxime1907:
Fixes the missing fields in schema and values that are present in `_helpers.tpl`:
- `fullnameOverride`
- `nameOverride`

### DIFF
--- a/resources/helm/dask-gateway/values.schema.yaml
+++ b/resources/helm/dask-gateway/values.schema.yaml
@@ -16,6 +16,16 @@ required:
   - rbac
   - global
 properties:
+  fullnameOverride:
+    type: [string, "null"]
+    description: |
+      fullnameOverride and nameOverride allow you to adjust how the resources
+      part of the Helm chart are named.
+  nameOverride:
+    type: [string, "null"]
+    description: |
+      See the description of fullnameOverride.
+
   gateway:
     type: object
     additionalProperties: false
--- a/resources/helm/dask-gateway/values.yaml
+++ b/resources/helm/dask-gateway/values.yaml
@@ -1,3 +1,11 @@
+## Provide a name to partially substitute for the full names of resources (will maintain the release name)
+##
+nameOverride: ""
+
+## Provide a name to substitute for the full names of resources
+##
+fullnameOverride: ""
+
 # gateway nested config relates to the api Pod and the dask-gateway-server
 # running within it, the k8s Service exposing it, as well as the schedulers
 # (gateway.backend.scheduler) and workers gateway.backend.worker) created by the


maxime1907:
cc @consideRatio

consideRatio:
Can you verify that the functionality described matches the behavior in this helm chart? This is copy pasted from jupyterhub/jupyterhub and some notes are related to that still, and the behavior in that helm chart is not standard as there were adjustments made to avoid introducing breaking changes when adding support for fullnameOverride and nameOverride.

- [ ] Validate documented claims
- [ ] Remove text about user pods that clearly is unrelated and review the notes in general, they probably don't make sense.

Alternatively, omit the descriptions. What is clear is that the schema should include the entries allowing them to be set, so just adding schemas without description in the schema is an improvement. I just don't want anything added to be misleading.

resources/helm/dask-gateway/values.yaml
```diff
@@ -1,3 +1,6 @@
+fullnameOverride: ""
+nameOverride: ""
```
consideRatio, line 2: To provide these default values as blank strings vs null values - what makes sense here? Are they both resulting in the same outcomes currently?

```yaml
fullnameOverride:
nameOverride:
```

consideRatio: I tried this practically, there was no difference.

Due to a bug in `helm` since 3.2.0+ subchart can't set something to null, let's go for a default value of null instead of a blank string for now as that gives us a bit more flexibility long term.

```suggestion
fullnameOverride:
nameOverride:
```

resources/helm/dask-gateway/values.schema.yaml
```diff
@@ -16,6 +16,16 @@ required:
...
 properties:
+  fullnameOverride:
+    type: [string, "null"]
+    description: |
+      fullnameOverride and nameOverride allow you to adjust how the resources
+      part of the Helm chart are named.
+  nameOverride:
+    type: [string, "null"]
+    description: |
+      See the documentation under [`fullnameOverride`](schema_fullnameOverride).
```
consideRatio, line 27:
```suggestion
      See the description offullnameOverride.
```

I'm removing the technical parts that only works when rendering these descriptions in Sphinx as done in jupyterhub/zero-to-jupyterhub-k8s but not done in this project.

resources/helm/dask-gateway/values.yaml
maxime1907: There is no difference between both but that's what most helm charts are doing nowadays:
- https://github.com/bitnami/charts/blob/master/bitnami/etcd/values.yaml#L31
- https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_names.tpl#L21
- https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L23

resources/helm/dask-gateway/values.schema.yaml
maxime1907: ah okay i see nice

resources/helm/dask-gateway/values.yaml
maxime1907: What do you think about trying to keep these values set to default values so that we follow what most people are doing ?
(I have rarely seen values set to null in helm charts)

consideRatio: Okay sure!

consideRatio:
Thanks @maxime1907!!

PR #606, opened 2022-09-19
Helm chart: add gateway.backend.imagePullSecrets

### LINKED ISSUE #584: Helm Chart: gateway.backend Missing "imagePullSecrets"
I noticed that the `gateway.backend` Helm chart schema was missing `imagePullSecrets` whilst attempting to use a custom image for workers/schedulers via `dask-gateway`.

I've taken a shot at addressing this addition in #583

maxime1907:
# Proposal

Currently, dask-gateway retrieves images that are hosted on public registries.
This behavior is fine but if we want to use private registries, it is only possible through `worker_extra_pod_config` and `scheduler_extra_pod_config`.
I have then tought of a way to implement it properly without introducing any breaking change.

# Implementation
- Add a `KubeClusterConfig` configuration field called `image_pull_secrets` that lets dask scheduler and worker pods to pull images from private registries.
- Change the helm chart to add support for the configuration field called `image_pull_secrets` thanks to the following field:
  - `imagePullSecrets`: Image pull secrets to access private registries

# Behind the scene
From the user and dask operator POV, it doesn't have any impact on the current behavior of dask-gateway.

---

- EDIT by consideRatio: closes #583 as this is a duplicate PR.
- EDIT by consideRatio: closes #584

### DIFF
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -56,7 +56,12 @@ class KubeClusterConfig(ClusterConfig):
 
     image_pull_secrets = List(
         [],
-        help="Image pull secrets to access private registries",
+        help="""
+        Image pull secrets to access private registries.
+        
+        Should be a list of dictionaries with a single key called ``name`` to
+        match the k8s native syntax.
+        """,
         config=True,
     )
 
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 
 from kubernetes_asyncio import client, config
 from kubernetes_asyncio.client.rest import ApiException
-from traitlets import Dict, Float, Unicode, default
+from traitlets import Dict, Float, List, Unicode, default
 from traitlets.config import LoggingConfigurable
 
 from ... import __version__ as VERSION
@@ -51,6 +51,12 @@ class KubeClusterConfig(ClusterConfig):
     image_pull_policy = Unicode(
         "IfNotPresent",
         help="The image pull policy of the docker image specified in ``image``",
+        config=True,
+    )
+
+    image_pull_secrets = List(
+        [],
+        help="Image pull secrets to access private registries",
         config=True,
     )
 
--- a/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
+++ b/dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
@@ -58,7 +58,7 @@ class KubeClusterConfig(ClusterConfig):
         [],
         help="""
         Image pull secrets to access private registries.
-        
+
         Should be a list of dictionaries with a single key called ``name`` to
         match the k8s native syntax.
         """,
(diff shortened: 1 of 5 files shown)


maxime1907:
cc @consideRatio

resources/helm/dask-gateway/values.yaml
```diff
@@ -107,6 +107,9 @@ gateway:
       tag: "set-by-chartpress"
       pullPolicy: IfNotPresent
 
+    # Image pull secrets for backend pod
```
consideRatio, line 110:
```suggestion
    # Image pull secrets for a dask cluster's scheduler and worker pods
```

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
```diff
@@ -1175,6 +1175,10 @@ def make_pod(self, namespace, cluster_name, config, is_worker=False):
             },
         }
 
+        pod["spec"]["imagePullSecrets"] = []
+        for image_pull_secret in config.image_pull_secrets:
+            pod["spec"]["imagePullSecrets"].append({"name": image_pull_secret})
+
```
consideRatio, line 1181: - [ ] Make this conditional, so that providing an empty list doesn't make the pod specification cluttered?
- [ ] Make this a straight passthrough of k8s native specification of imagePullSecrets, don't make it accept the names of k8s secrets but accept a list of `{"name": "secret-name-here"}`.

  I'd like that as it aligns with the helm chart's other imagePullSecrets specification. See for example:
  https://github.com/dask/dask-gateway/blob/a660e2a4820a51e4027294bbd6ff8ed17e477563/resources/helm/dask-gateway/templates/gateway/deployment.yaml#L37-L40

consideRatio (changes requested):

consideRatio:
Also, I see that there is descriptions in the schema file about `tag, pullPolicy, or pullSecrets`, could you delete the `pullSecrets` part as that is not correct and semi-related to this PR that implements the `imagePullSecrets` field?

dask-gateway-server/dask_gateway_server/backends/kubernetes/controller.py
consideRatio:
```suggestion
        if config.image_pull_secrets:
            pod["spec"]["imagePullSecrets"] = config.image_pull_secrets

```

dask-gateway-server/dask_gateway_server/backends/kubernetes/backend.py
```diff
@@ -54,6 +54,12 @@ def _default_namespace(self):
         config=True,
     )
 
+    image_pull_secrets = List(
+        [],
+        help="Image pull secrets to access private registries",
```
consideRatio, line 59:
```suggestion
        help="""
        Image pull secrets to access private registries.
        
        Should be a list of dictionaries with a single key called ``name`` to
        match the k8s native syntax.
        """,
```

consideRatio (approved):
With the suggested changes, this LGTM.

consideRatio:
@maxime1907 I applied the change suggestions, if this looks good to you, let's go for a merge!

maxime1907:
> @maxime1907 I applied the change suggestions, if this looks good to you, let's go for a merge!

Thank you a lot!!
Yes you can merge like this 🙂

consideRatio:
Thank you @maxime1907, @chadsr, and @szachovy for your contributions towards this!!! :heart: :tada: :sunflower:

PR #609, opened 2022-09-20
Add AttributeError exception for _ipython_display_

### LINKED ISSUE #607: 'VBox' object has no attribute '_ipython_display_'
**What happened**:

Received `AttributeError: 'VBox' object has no attribute '_ipython_display_'` when calling `cluster_options` from within a Jupyter notebook.
image (1)

**What you expected to happen**:

The options widget should be shown correctly, as expected.

**Minimal Complete Verifiable Example**:

```python
from dask_gateway import Gateway

gt = Gateway()
options = gt.cluster_options()
options
```

**Anything else we need to know?**:

**Environment**:

- dask-gateway: 2022.4.0
- Dask version: 2022.4.2
- Python version: 3.9.13
- Operating System: Linux
- Install method (conda, pip, source): conda

## Extra

The reason for the above issue seems to be that Ipywidgets latest available version `8.0.0` removed the `on_display` method in place of `_repr_mimebundle_` (see [ref.](https://github.com/jupyter-widgets/ipywidgets/pull/2021)) and the Dask-gateway client still uses the old call: 
https://github.com/dask/dask-gateway/blob/376b4a29622ed6cdb0fe9707d589ec625be5d079/dask-gateway/dask_gateway/client.py#L1222-L1230

viniciusdc:
Fixes #607 

After `ipywidgets` released their `8.0.0` version the `_ipython_display_` method was renamed to `_repr_mimebundle_`, this handles the method's incompatibility by using an AttributeError exception.

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -1222,7 +1222,11 @@ class GatewayCluster:
     def _ipython_display_(self, **kwargs):
         widget = self._widget()
         if widget is not None:
-            return widget._ipython_display_(**kwargs)
+            # ipywidgets renamed the _ipython_display_ method in 8.x
+            try:
+                return widget._ipython_display_(**kwargs)
+            except AttributeError:
+                return widget._repr_mimebundle_(**kwargs)
         else:
             from IPython.display import display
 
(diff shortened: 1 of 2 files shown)


consideRatio (approved):
Wiee thanks for effort to debug, resolve, and leave a relevant inline comment to ensure we can clean this up in the future.

PR #611, opened 2022-10-07
refactor: use traefik documented syntax for cli flags

consideRatio:
This PR is a simple refactoring to use CLI flags for traefik that matches what they document they accept.

This seem to work no matter what, but when debugging things earlier it made me wonder if we had an issue passing these flags as their case didn't align with the documentation in traefik. This refactoring makes us align with the documented way of passing the flags.


martindurant (approved):

consideRatio:
Thanks @martindurant!

PR #612, opened 2022-10-07
Simplified integration with namespace local JupyterHub Helm charts

### LINKED ISSUE #473: Helm chart: allow JUPYTERHUB_API_KEY to be mounted from specified k8s secret and key
### Feature proposal summarized

I propose allowing an environment variable for the gateway pod to be set by not passing an API key directly, but by passing a name for a k8s Secret resource and a key to read a value from within it. This is helpful as the JupyterHub helm chart for example supports automatic generation of JupyterHub API keys to be stored in a k8s Secret - so then one could simply reference that without asking the helm chart user to manage the sensitive API key at any point in time.

Practically, something like this is what is proposed: allowing an optional configuration to be used.

```diff
    {{- if (eq .Values.gateway.auth.type "jupyterhub") }} 
    - name: JUPYTERHUB_API_TOKEN 
      valueFrom: 
        secretKeyRef: 
+         name: {{ .Values.gateway.auth.jupyterhub.apiTokenFromSecretName | default (include "dask-gateway.apiName" .) }} 
+         key: {{ .Values.gateway.auth.jupyterhub.apiTokenFromSecretKey | default "jupyterhub-api-token" }} 
-         name: {{ include "dask-gateway.apiName" . }} 
-         key: jupyterhub-api-token 
    {{- end }} 
```

### A technical background

[`jupyterhub_api_token` and `jupyterhub_api_url`](https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/dask-gateway-server/dask_gateway_server/auth.py#L284-L310) is configuration for the `JupyterHubAuthenticator` class. The dask-gateway Helm chart [sets the JUPYTERHUB_API_TOKEN environment variable via a k8s Secret](https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/resources/helm/dask-gateway/templates/gateway/deployment.yaml#L56-L63) that in turn [is passed a value from helm chart config](https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/resources/helm/dask-gateway/templates/gateway/secret.yaml#L9-L10).

In Helm chart's like `daskhub` that deploys a jupyterhub and dask-gateway next to each other in the same k8s namespace, you are [supposed to declare an api token in the dask-gateway helm chart config, and the jupyterhub helm chart config](https://github.com/dask/helm-chart/tree/main/daskhub#prepare-configuration-file) like this:

```yaml
# a values.yaml file passed when using daskhub that
# depends on the jupyterhub and dask-gateway helm chart
jupyterhub:
  hub:
    services:
      dask-gateway:
        apiToken: "<secret-token>"

dask-gateway:
  gateway:
    auth:
      jupyterhub:
        apiToken: "<secret-token>"
```

The JupyterHub Helm chart can now automatically generate an API token for any services under `hub.services` and store it in a k8s Secret. If the dask-gateway helm chart could be made to reference that k8s Secret, it could mount it from there directly instead of first creating its own k8s Secret with passed api token, and then mounting it.

Currently though, there is no mechanism to provide a custom environment variable to the dask-gateway-server pod, or configure it to mount the JUPYTERHUB_API_TOKEN environment variable from another k8s Secret.

https://github.com/dask/dask-gateway/blob/bee9255e5ea0d77f456985cd91b2622bb3776dbb/resources/helm/dask-gateway/templates/gateway/deployment.yaml#L56-L63

I think a clean solution would be to allow this environment variable to be mounted from a manually specified k8s secret and key. That way, the daskhub helm chart for example, could simply specify this instead...

```yaml
# possible future default values for a future version of daskhub that
# depends on the jupyterhub and dask-gateway helm chart, where
# the dask-gateway helm chart allows configuration of what k8s Secret
# and key it use when mounting a JUPYTERHUB_API_TOKEN
# environment variable to the dask-gateway-server pod
#
# like this, users of the helm chart won't need to generate or maintain
# a need-to-keep secret jupyterhub api token
jupyterhub:
  hub:
    services:
      # By having an entry for dask-gateway, the z2jh helm chart will
      # generate a jupyterhub api token and persist it in the hub k8s Secret
      dask-gateway: {}

dask-gateway:
  gateway:
    auth:
      jupyterhub:
        apiTokenFromSecretName: hub
        apiTokenFromSecretKey: hub.services.dask-gateway.apiToken
```

### Implementation idea

1. Allow a k8s Secret name and key be configurable to mount the JupyterHub API token from a custom k8s Secret with a given key.
2. Make the creation of the dask-gateway managed k8s Secret be conditional of not providing a custom k8s Secret name / key.
3. Update the values schema with these settings

consideRatio:
## Summary

Closes #473. The idea is to enable users installing both the dask-gateway chart and jupyterhub chart to not provide any api-token credential for dask-gateway/juyterhub to trust each other, but instead rely on a generated api-token persisted in a k8s Secret.

The implementation plan in #473 was the following:

> ### Implementation idea
> 1. Allow a k8s Secret name and key be configurable to mount the JupyterHub API token from a custom k8s Secret with a given key.
> 2. Make the creation of the dask-gateway managed k8s Secret be conditional of not providing a custom k8s Secret name / key.
> 3. Update the values schema with these settings

This PR follows that plan quite well, but provides a default for the k8s Secret name and Key config and relies on them by default unless the previously required `apiToken` is specified. This PR also updates the documentation under the topic of installing the dask-gateway helm chart and autenticating against a JupyterHub Helm chart installation.

## Added chart config in values.yaml

```yaml
gateway:
  auth:
    jupyterhub:
      # The JupyterHub Helm chart will automatically generate a token for a
      # registered service. If you don't specify an apiToken explicitly as
      # required in dask-gateway version <=2022.6.1, the dask-gateway Helm chart
      # will try to look for a token from a k8s Secret created by the JupyterHub
      # Helm chart in the same namespace. A failure to find this k8s Secret and
      # key will cause a MountFailure for when the api-dask-gateway pod is
      # starting.
      apiTokenFromSecretName: hub
      apiTokenFromSecretKey: hub.services.dask-gateway.apiToken
```

The schema file is also updated to reflect this new configuration.

## Successfully tested

I've tested this on the hub.jupytearth.org deployment of a dask-gateway deployed next to a jupyterhub.

## How dask/helm-chart's daskhub would adjust with this

[Steps relating to generating and configuring an api token](https://github.com/dask/helm-chart/blob/a9f94ca7a040b4a42c0554ebb4f9bf55c2ce7ec7/daskhub/.frigate#L26-L56) could be simplied. This would be an example of the configuration to have, without secret credentials involved.

```yaml
jupyterhub:
  hub:
    services:
      dask-gateway:
        display: false

dask-gateway:
  gateway:
    auth:
      type: jupyterhub
```

### DIFF
--- a/docs/source/install-kube.rst
+++ b/docs/source/install-kube.rst
@@ -300,36 +300,70 @@ JupyterHub_ provides a multi-user interactive notebook_ environment. Through
 the zero-to-jupyterhub-k8s_ project, many companies and institutions have setup
 JuypterHub to run on Kubernetes. When deploying Dask-Gateway alongside
 JupyterHub, you can configure Dask-Gateway to use JupyterHub for
-authentication. To do this, we register ``dask-gateway`` as a `JupyterHub
-Service`_.
+authentication.
 
-First we need to generate an API Token - this is commonly done using
-``openssl``:
+Configuring a dask-gateway chart with a jupyterhub chart is more straight
+forward if they are installed in the same namespace for two reasons. First the
+JupyterHub chart generates api tokens for registered services and puts them in a
+k8s Secret that dask-gateway can make use of. Secondly dask-gateway
+pods/containers can detect the k8s Service from the JupyterHub chart's resources
+in the automatically.
+
+If dask-gateway **is installed in the same namespace as jupyterhub**, this is the
+recommended configuration to use.
+
+.. code-block:: yaml
+
+   # jupyterhub chart configuration
+   hub:
+     services:
+       dask-gateway:
+         display: false
+
+.. note::
+
+   The ``display`` attribute hides dask-gateway from the 'Services' dropdown in
+   the JupyterHub home page as dask-gateway doesn't offer any UI.
+
+.. code-block:: yaml
+
+   # dask-gateway chart configuration
+   gateway:
+     auth:
+       type: jupyterhub
+
+.. note::
+
+   This configuration relies on the dask-gateway chart's default values of
+   ``display.auth.jupyterhub.apiTokenFromSecretName`` and
+   ``display.auth.jupyterhub.apiTokenFromSecretKey`` as can be inspected in the
+   `default values.yaml file`_.
+
+If dask-gateway **isn't installed in the same namespace as jupyterhub**, this is
+the recommended configuration procedure.
+
+First generate an api token to use, for example using using ``openssl``:
 
 .. code-block:: shell
 
    openssl rand -hex 32
 
-Then add the following lines to your ``config.yaml`` file:
+Once you have it, your configuration should look like below, where ``<API URL>``
+should look like ``https://<JUPYTERHUB-HOST>:<JUPYTERHUB-PORT>/hub/api`` and
+``<API TOKEN>`` should be the generated api token.
 
 .. code-block:: yaml
 
-   gateway:
-     auth:
-       type: jupyterhub
-       jupyterhub:
+   # jupyterhub chart configuration
+   hub:
+     services:
+       dask-gateway:
          apiToken: "<API TOKEN>"
-
-replacing ``<API TOKEN>`` with the output from above.
-
-If you're not deploying Dask-Gateway in the same cluster and namespace as
-JupyterHub, you'll also need to specify JupyterHub's API url. This is usually
-of the form ``https://<JUPYTERHUB-HOST>:<JUPYTERHUB-PORT>/hub/api``. If
-JupyterHub and Dask-Gateway are on the same cluster and namespace you can omit
-this configuration key, the address will be inferred automatically.
+         display: false
 
 .. code-block:: yaml
 
+   # dask-gateway chart configuration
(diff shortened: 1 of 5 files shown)


TomAugspurger (approved):
Very cool @consideRatio!

One quick question: do you worry at all about upgrades to existing deployments? I suspect things will be just fine, since providing your own tokens continues to be an option.

consideRatio:
> One quick question: do you worry at all about upgrades to existing deployments? I suspect things will be just fine, since providing your own tokens continues to be an option.

Exactly, previously everyone has been forced to provide a token - so they should all be fine as this will only impact people not providing a token.

TomAugspurger:
Great, thanks. Feel free to merge whenever you're ready!

consideRatio:
Thanks! Going for it!

PR #620, opened 2022-10-13
ci: test against latest versions of k8s

consideRatio:


martindurant:
Good idea! Is there a reason to want to test against multiple versions of kubernetes?

consideRatio:
> Is there a reason to want to test against multiple versions of kubernetes?

Yepp! The Helm chart that we test is compatible with some versions of k8s, but not all, because of resources evolve over time.

By testing against the oldest known supported version of k8s and newest version of k8s, we ensure people installing the Helm chart and upgrading to newer versions of k8s won't run into issues.

The kind of changes made over time are often quite simple - but essential for function. Here is an example of such maintenance: https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/920

.github/workflows/test.yaml
```diff
@@ -129,8 +129,8 @@ jobs:
           - k3s-channel: v1.20
             upgrade-from: "0.9.0"
           - k3s-channel: v1.21
```
consideRatio, line 131: This can safely be removed I think.

```suggestion
```

PR #621, opened 2022-10-13
helm chart: update traefik to 2.9.1 and the associated CRDs

consideRatio:
I don't know that doing this will be relevant, but it was a while since traefik was updated and it seems like good hygine to keep it updated somewhat regularly.


consideRatio:
Merging this in case its relevant for #626 to have a modern Traefik version.

Issue #250, opened 2020-04-14
Support mirroring of secrets into the gatway server namespace


jcrist:
Some users have brought up qualms about our current RBAC requirements (specifically around secrets). Currently we require:

- Controller: `create`, `delete` in all namespaces
- API Server: `get` in all namespaces
- Traefik: `get`, `list`, `watch` in all namespaces

The worry is around giving the ability to read secrets in all namespaces to pods. This affects the API Server and the Traefik proxy pods.

We don't rely on secrets for Traefik, but the implementation in Traefik gets grumpy if you disable this RBAC permission. An upstream fix would be needed, and I'm not likely to do that (if others do, please let me know and we can update our chart).

We change the requirements for the API server though. While a dask cluster is active, the controller would ensure there was a mirrored copy of the secret in the same namespace as the API server. The API server would then only need `get` permissions for secrets in its own namespace. This adds some logic complications in the controller. The mirrored secrets also wouldn't have an parent objects, so their deletion would have to be handled explicitly in the controller.

I'm hesitant to make this change (as it complicates the codebase and uses more resources), but could if needed. If possible I'd like to make this optional, defaulting to the existing logic (no mirroring).

jcrist:
cc @gforsyth, @dankerrigan, @mmccarty

droctothorpe:
One alternative to consider is an aggregate role, but it would only be applicable in a double-namespace (as opposed to a dynamic namespace) deployment pattern.

I relate to your reservations, @jcrist, and I'm not sure it's worth the added complexity if other widely adopted tools, like Traefik, consider the ability to `get` but not `list` secrets in all namespaces an acceptable paradigm. 

Here are examples of two other widely adopted operators that use the same model:
*  [Spark Operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/manifest/spark-operator-rbac.yaml#L40)
* [ArgoCD Operator](https://github.com/argoproj-labs/argocd-operator/blob/e540fb87289969e76c95dbba2927d71161778f02/deploy/role.yaml#L254)

I think that this isn't an anti-pattern. As long as the controller can't `list`, we're leveraging security through obscurity.

shaunc:
[...]

PR #626, opened 2022-10-20
Reduce RBAC privileges for single namespace deployment

holzman:
## Overview

The dask-gateway helm chart deploys three microservices: gateway, controller, traefik into a single namespace.   

When a user requests a new dask cluster, a number of Kubernetes custom resource definition objects are created, as well as pods for dask-scheduler and dask-workers.   By default, these are deployed into the same namespace as the microservices, but can be configured to deploy to a separate namespace via the `gateway.backend.namespace` helm value.

For the same-namespace deployment, the chart creates cluster-wide permissions that are unnecessarily permissive.  This PR checks if the user has set `gateway.backend.namespace` to a different namespace.  In the case of a single-namespace deployment, it moves `ClusterRole` permissions to namespaced `Role` permissions
whenever possible.   It also adds an option to the traefik chart so that it only listens on a single namespace.

In addition, for the single-namespace deployment, the code itself is also unnecessarily broad - for example, it performs cluster-wide queries for `daskcluster` when it should restrict itself to namespaced queries.  This PR does *not* address this, but https://github.com/dask/dask-gateway/pull/594 does.

#### What I wrote previously: 

If we are deploying into a single namespace (gateway.backend.Namespace is either null or the same as the release namespace), move as many privileges as possible from cluster-wide to namespaced ones.

This is one possible mitigation to the security concerns discussed in #250.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -128,8 +128,13 @@ jobs:
         include:
           - k3s-channel: v1.20
             upgrade-from: "0.9.0"
+            dask-namespace: "default"
           - k3s-channel: stable
+            dask-namespace: "default"
+          - k3s-channel: stable
+            dask-namespace: "additional"
           - k3s-channel: latest
+            dask-namespace: "default"
 
     steps:
       - uses: actions/checkout@v3
@@ -172,6 +177,10 @@ jobs:
               --include-crds \
               --values=resources/helm/testing/chart-install-values.yaml
 
+      - if: matrix.dask-namespace == 'additional'
+        run: |
+          kubectl create namespace additional
+
       - name: helm install previous version ${{ matrix.upgrade-from }}
         if: matrix.upgrade-from != ''
         run: |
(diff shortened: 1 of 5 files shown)


consideRatio:
@holzman we used traefik 2.6.x, now 2.9.x if you rebase this PR on `main`. Just wanted to mention that in case you would run into a bug fixed already in traefik.

consideRatio:
@holzman I'm low on maintenance capacity, but note that #594 is open as well - I think you look to accomplish similar things. It would be good if you help coordinate with that PR in some way.

holzman:
Hi Erik - 

I was going to propose something similar to #594 as a follow-up PR - this PR only changes the helm chart and doesn't change any dask-gateway code.

I can certainly integrate this with #594 if you'd prefer to consolidate.

consideRatio:
Help me and other maintainers overview the changes as best as you can in the PR description, if this should include other changes or not is hard for me to evaluate without the overview.

holzman:
Ok - I'll fix up the code (tests are failing because it's missing some privileges for the multi-namespace case) and write up a clear overview for this PR.

holzman:
Ah, this exposed an issue with the current tests - they assume that the scheduler/worker namespace is "default" and don't cover a test case where the scheduler/worker namespace is different.

I'll submit a clean separate PR that only extends the test coverage.

consideRatio:
Closed temporarily as motivated in https://github.com/dask/dask-gateway/pull/594#issuecomment-1361333469

PR #634, opened 2022-11-04
Fix failure to build and publish arm64 images

### LINKED ISSUE #629: Build arm64 dask-gateway-server image
Currently, the images on docker hub only support linux/amd64, but our cluster is primarily arm64 - infact the only x86 pods are for dask-gateway-server.

I have built the image locally and can confirm that it builds and runs without issue.

consideRatio:
## Summary

- A bugfix of dask-gateway-server image for arm64 wasn't built or published
- A new feature of providing the dask-gateway image (demo for scheduler/workers) for arm64 as well. Now the entire helm chart is arm64 compatible.
  - This involved adding a build stage where compilers etc were available to build wheels (psutils specifically), and then using them in the final stage of the Dockerfile. Note there is a quite cool trick involved here, were we avoid a COPY operation, but instead using `--mount` for the `RUN` command, making things from the previous stage accessible without creating another layer that takes up space in the final image.

## Background

There is a [bug in `chartpress`](https://github.com/jupyterhub/chartpress/pull/193) making us fail to build and publish the arm64 variant of the dask-gateway-server image. This PR makes us avoid that bug by making the dask-gateway image also support arm64.

- Closes #629

### DIFF
--- a/dask-gateway/Dockerfile
+++ b/dask-gateway/Dockerfile
@@ -11,7 +11,36 @@
 #
 # See https://gateway.dask.org/install-kube.html#using-a-custom-image.
 #
-FROM python:3.10-slim-bullseye
+
+
+# The build stage
+# ---------------
+# This stage is building Python wheels for use in later stages by using a base
+# image that has more pre-requisites to do so, such as a C++ compiler.
+#
+# psutils, a dependency of distributed, is currently the sole reason we have to
+# have this build stage.
+#
+FROM python:3.10-bullseye as build-stage
+
+# Build wheels
+#
+# We set pip's cache directory and expose it across build stages via an
+# ephemeral docker cache (--mount=type=cache,target=${PIP_CACHE_DIR}).
+#
+COPY . /opt/dask-gateway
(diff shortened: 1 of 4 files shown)


dask-gateway/Dockerfile
```diff
@@ -34,7 +63,11 @@ WORKDIR /home/dask/
 
 # Install dask-gateway
 COPY --chown=dask:dask . /opt/dask-gateway
-RUN pip install --no-cache-dir \
+ARG PIP_CACHE_DIR=/tmp/pip-cache
```
jacobtomlinson, line 66: Does this need to be set a second time?

jacobtomlinson (approved):
Neat!

dask-gateway/Dockerfile
consideRatio: Yepp, another stage, another built-time env var needed to configure pip where to store/read cache

consideRatio:
Going for a self-merge and a changelog for review/merge -> release

PR #636, opened 2022-11-07
helm chart: update traefik to 2.9.1 (associated CRDs unchanged)

consideRatio:


consideRatio:
I tried to test against python 3.11 etc, but ran into issues. Ignoring failures in pbs/slurm/hadoop for now.

PR #644, opened 2022-11-11
ci: use ubuntu-22.04 explicitly and constrain test duration

consideRatio:


consideRatio:
Hadoop / PBS tests seem to get stuck from time to time for some reason.

PR #672, opened 2023-01-10
client: fix failure to display widget with ipywidgets 8

### LINKED ISSUE #671: Widget display is missing for ipywidget>=8
**Describe the issue**:

When using DaskGateway in jupyterlab with `ipywidgets==8.0.4`, `._ipython_display_()` method of `Options` and `DaskGatewayCluster` does not actually display anything.

 https://github.com/dask/dask-gateway/blob/b585775fd8d948006f67846bc96e904e38e2e114/dask-gateway/dask_gateway/options.py#L76-L81

From IPython docs:

> \_ipython_display\_()
> Displays the object as a side effect; the return value is ignored. If this is defined, all other display methods are ignored.

So one has to display things rather than return a model, this is not happening for newer versions of `ipywidgets`.

**Minimal Complete Verifiable Example**:

Run this in jupyterlab:

```python
from IPython.display import display
from dask_gateway.options import Options
opts = Options()
display(opts) # < nothing is displayed
display(opts._widget()) # < widget is displayed
```

Same issue in `DaskGatewayCluster`

https://github.com/dask/dask-gateway/blob/b585775fd8d948006f67846bc96e904e38e2e114/dask-gateway/dask_gateway/client.py#L1225-L1230

**Anything else we need to know?**:

**Environment**:

`dask_gateway==2022.10.0` from conda-forge

- Dask version: 2022.12.1
- Python version: 3.10.8
- Operating System: Linux
- Install method (conda, pip, source): conda

Kirill888:
Rather than calling `widget._ipython_display_()` use `display(widget)`, this should support older and newer versions of `ipywidget`.

Closes #671

### DIFF
--- a/dask-gateway/dask_gateway/client.py
+++ b/dask-gateway/dask_gateway/client.py
@@ -1221,16 +1221,12 @@ class GatewayCluster:
         return box
 
     def _ipython_display_(self, **kwargs):
+        from IPython.display import display
+
         widget = self._widget()
         if widget is not None:
-            # ipywidgets renamed the _ipython_display_ method in 8.x
-            try:
-                return widget._ipython_display_(**kwargs)
-            except AttributeError:
-                return widget._repr_mimebundle_(**kwargs)
+            display(widget, **kwargs)
         else:
-            from IPython.display import display
-
             data = {"text/plain": repr(self), "text/html": self._repr_html_()}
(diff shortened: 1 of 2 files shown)


consideRatio:
Thank you @Kirill888!!! :heart: :tada: :sunflower:

PR #673, opened 2023-01-10
Update RELEASE.md to use tbump

consideRatio:
From manually needing to update, commit, tag, and push three files with version fields - we now do it all with a tbump command. This command will also summarize clearly what it will do via `tbump --dry-run <version>` as also suggested in the REALEASE.md file before one does it without --dry-run.

And, the config on what files to update etc are nicely located into a pyproject.toml file.


consideRatio:
I'll go for a merge and create a new release to get #672 out.

PR #677, opened 2023-01-23
client: fix double display of Options in jupyter

Kirill888:
This was missed in #672. After displaying options as a widget, textual representation is also displayed.


consideRatio (approved):

consideRatio:
Thank you @Kirill888!!!

PR #688, opened 2023-02-27
Adding envs key Helm values to gateway resources

JoeJasinski:
This MR is to enhance this use-case where one wishes to customize the extraConfig that's passed to the gateway via the configmap.  Currently, [according to the documentation](https://gateway.dask.org/install-kube.html#using-extraconfig), you can create additional python configuration for any of the Traitlet-enhanced classes from the "[Configuration Reference"](https://gateway.dask.org/api-server.html). 

However, it would be nice to be able to pass in environment variables to that config, so you could set the python config additions, but tune them by making a change to an environment variable rather than via making python code changes. 

Take this simple Helm values file example: 

```yaml
gateway:
  extraConfig:
    # Note that the key name here doesn't matter. Values in the
    # `extraConfig` map are concatenated, sorted by key name.
    clusteroptions: |
      c.KubeClusterConfig.idle_timeout = 300
      c.KubeClusterConfig.cluster_max_memory = "20G"
```
If I wanted to customize the config values, I'd have to override the entire config, or create a new "extraConfig" key.

However, it would be nice to be able to do the following:

```yaml
gateway:
  envs:
  - name: IDLE_TIMEOUT
    value: "500"
  extraConfig:
    # Note that the key name here doesn't matter. Values in the
    # `extraConfig` map are concatenated, sorted by key name.
    clusteroptions: |
      import os
      idle_timeout =  int(os.environ.get("IDLE_TIMEOUT", 300))
      cluster_max_memory = os.environ.get("CLUSTER_MAX_MEMORY", "20G")
      c.KubeClusterConfig.idle_timeout = idle_timeout
      c.KubeClusterConfig.cluster_max_memory = cluster_max_memory

```

Then, you could easily change those config values by passing in environment variables to the helm chart without having to change the python code. 

This MR is to change the helm chart so that it accepts an "envs" list of dicts that get passed to the gateway deployment where the dask gateway config ConfigMap gets passed into the gateway. 

Does this sound like a feature that would be useful?

### DIFF
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -54,9 +54,8 @@ spec:
             - mountPath: /etc/dask-gateway/
               name: configmap
           env:
-            {{- range .Values.gateway.env }}
-            - name: {{ .name }}
-              value: {{ .value| quote }}
+            {{- with .Values.gateway.env }}
+            {{- . | toYaml | nindent 12 }}
             {{- end }}
             {{- if (eq .Values.gateway.auth.type "jupyterhub") }}
             - name: JUPYTERHUB_API_TOKEN
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -54,7 +54,7 @@ spec:
             - mountPath: /etc/dask-gateway/
               name: configmap
           env:
-            {{- range .Values.controller.env }}
+            {{- range .Values.gateway.env }}
             - name: {{ .name }}
               value: {{ .value| quote }}
             {{- end }}
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -42,6 +42,11 @@ spec:
         - name: gateway
           image: {{ .Values.gateway.image.name }}:{{ .Values.gateway.image.tag }}
           imagePullPolicy: {{ .Values.gateway.image.pullPolicy }}
+          env:
+          {{- range $key, $value := .Values.gateway.envs }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+          {{- end }}
           args:
             - dask-gateway-server
             - --config
--- a/resources/helm/dask-gateway/templates/gateway/deployment.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/deployment.yaml
@@ -42,11 +42,6 @@ spec:
         - name: gateway
           image: {{ .Values.gateway.image.name }}:{{ .Values.gateway.image.tag }}
           imagePullPolicy: {{ .Values.gateway.image.pullPolicy }}
-          env:
-          {{- range $key, $value := .Values.gateway.envs }}
-            - name: {{ $key }}
-              value: {{ $value | quote }}
-          {{- end }}
           args:
             - dask-gateway-server
             - --config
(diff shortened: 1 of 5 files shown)


resources/helm/dask-gateway/templates/gateway/deployment.yaml
```diff
@@ -54,6 +54,10 @@ spec:
             - mountPath: /etc/dask-gateway/
               name: configmap
           env:
+            {{- range .Values.gateway.env }}
+            - name: {{ .name }}
+              value: {{ .value| quote }}
+            {{- end }}
```
consideRatio, line 60: Like this, a user can provide any kind of `env` as examplified by for example https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/ or via https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#envvar-v1-core which is the definitive description of what should be allowed.

```suggestion
            {{- with .Values.gateway.env }}
            {{- . | toYaml | nindent 12 }}
            {{- end }}
```

resources/helm/dask-gateway/values.schema.yaml
```diff
@@ -108,6 +108,25 @@ properties:
...
+                The name of the environment variable to pass to the pod
+            value:
+              type: string
+              description: |
+                The value of the environment variable to pass to the pod
+          required:
+            - name
+            - value
+        description: |
+          Additional environment variables to apply to the resource
```
consideRatio, line 129: This should allow for configuration as declared in https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#envvar-v1-core, including `name` (string), `value` (string), and `valueFrom` (object).

I'd like to see that:
- the description mention that users should use the k8s official configuration syntax
- properties entries for name/value/valueFrom declaring them string/string/object, you can skip the descriptions here to defer to the k8s official docs instead
- for the valueFrom object, just declare `additionalProperties: true` and don't declare its properties
- remove the `&env-spec` anchor definition if its not used elsewhere
- Only `name` is required as otherwise you can't choose from either `value` or `valueFrom`.

resources/helm/testing/chart-install-values.yaml
```diff
@@ -1,6 +1,9 @@
 gateway:
   loglevel: DEBUG
   prefix: /services/dask-gateway
+  env:
+  - name: MY_ENV
+    value: My env
```
consideRatio, line 6: Add a test case with a `valueFrom` entry as well.

```suggestion
  env:
  - name: MY_ENV
    value: My env
  - name: MY_CPU_REQUESTS
    valueFrom:
      resourceFieldRef:
        containerName: gateway
        resource: requests.cpu
```

consideRatio (commented):
Thank you @JoeJasinski for your thorough descriptions in PR and issue!!! :heart: :tada: :sunflower: 

I'm :+1: for supporting `gateway.env` as a configuration, but I want it to be fully k8s compatible and support `valueFrom` for example.

I've made comments to help you work towards that!

JoeJasinski:
@consideRatio Thank you for the feedback. I like that approach of allowing the downward API access to set environment variables. I'll make these changes.

JoeJasinski:
@consideRatio Thank you again for the review. I believe I made the changes that you pointed out. I like the toYaml loop that you added for the environment variable spec; I'll have to include that in my Helm charts. I've updated the jsonschema to allow for the valuesFrom key. 

Let me know if there are any other updates you'd like me to make.

consideRatio (approved):

consideRatio:
Wieee this looks great! Thank you @JoeJasinski!! :heart: :tada: :sunflower:

PR #739, opened 2023-09-22
Require py39, k8s 1.24, use golang 1.21, update traefik from v2.9 to v2.10

consideRatio:
- Non-breaking: use golang 1.21 to compile dask-gateway-proxy
- Dropping support for Python 3.8 as that is what dask/distributed and other packages has done already in the dask ecosystem.
- Dropping support for k8s 1.23 and lower. For related discussions about things like this, see https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3042.
- Upgrading the helm chart's traefik version from v2.9 to v2.10, which forces users upgrading the helm chart to also upgrade bundled CRD resources. A note is to be made in the changelog about doing `kubectl apply` before ugprading as has been done for previous releases when this has become needed.


TomAugspurger (approved):

consideRatio:
Thanks for reviewing @TomAugspurger!

I did a retroactive self-review and realized a possible issue for existing clusters, but after reviewing it further I don't think its an issue - I believe existing clusters will continue to work well even if the dask-gateway helm chart is upgraded.

By adopting the some new stuff in traefik v2.10, the upgrade to traefik v3 at some time in the future will be less disruptive though - as clusters created with traefik v2.9 (but not v2.10) would need to be re-created before traefik v3 is used.

Without objections by tomorrow, I'll go for a merge and cut a release. As can be seen from https://github.com/dask/dask-gateway/commit/f2e0e4801c9c8287c18bc3e5113c33633bea846c, its not much that we've done, but there are some bugfixes and a new feature for the helm chart.

PR #748, opened 2023-09-28
Fix command applying CRDs from bad release in changelog

brews:
Hey! Noticed the [Breaking Changes section in the changelog for the 2023.9.0 release](https://gateway.dask.org/changelog.html#breaking-changes) has people manually update CRDs with `kubectl`. The problem is the command will force-apply the CRDs from the much older 2022.11.0 release. 

This PR updates the commands in the changelog so the command applies resources from the 2023.9.0 release.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -15,8 +15,8 @@
   below:
 
   ```shell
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio:
Thank you @brews !!!

PR #766, opened 2023-10-30
k8s controller: round to worker cores to three digits, not one

### LINKED ISSUE #765: Kubernetes controller deoesn't respect worker_cores factions correctly
The dask-gateway controller doesn't respect the `worker_cores` configuration properly. A DaskCluster specification with `worker_cores=7.745` ends up with a pod specification requesting `7700m`. It seems that it rounds to the nearest 100m CPU.

consideRatio:
I believe that CPU requests as enforced by the container runtimes associated with k8s clusters often has a resolution to millicores or 0.001 of a core, so going for three decimals makes us not loose resolution without introducing loads of decimals.

- Fixes #765


TomAugspurger (approved):

consideRatio:
Thank you @TomAugspurger!!

PR #770, opened 2023-11-17
No longer explicitly set imagePullPolicy to IfNotPresent by default

consideRatio:
This PR is inspired by https://github.com/jupyterhub/kubespawner/pull/807.

This PR influences the specific situation when a `latest` tagged image is to be used, then omiting the imagePullPolicy will result in a changed behavior to instead act as if imagePullPolicy was specified to `Always`. For any other image tag, omitting `imagePullPolicy` still behaves as if its set to `IfNotPresent`.


TomAugspurger (approved):
Makes sense, thanks.

consideRatio:
Thank you for reviewing @TomAugspurger!!

PR #794, opened 2024-01-04
Drop support for k8s 1.24, test against 1.29

consideRatio:
k8s 1.24 has reached EOL for various cloud providers, so we drop support for it now and require k8s 1.25+.

This is related to https://github.com/jupyterhub/zero-to-jupyterhub-k8s/issues/3309 that includes more details on cloud providers EOL for various k8s distributions.


consideRatio:
I'll go for a merge here and get a release out with this in #743, note that k8s 1.24 is currently very old even among cloud provider's.

PR #840, opened 2025-01-20
ci: run tests etc with ubuntu 24.04 up from 22.04

consideRatio:


consideRatio:
No additional breaking of CI system than is already broken it seems, going for it!

Issue #814, opened 2024-04-02
Tests broken again


consideRatio:

jacobtomlinson:
Thanks for taking the time here @consideRatio!

PR #841, opened 2025-01-20
Require python 3.10+, up from 3.9+

consideRatio:
I see dask/distributed requires 3.10+, so I figure we should reduce the complexity of dask-gateway maintenance by requiring that as well.

I'm doing this now since I'm trying to debug issues in test failures (https://github.com/dask/dask-gateway/issues/814), removing py39 which needs to use an older distributed version could reduce some complexity.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -56,14 +56,14 @@ jobs:
           #   we test integration with. We can test against only one version and
           #   that would be fine.
           #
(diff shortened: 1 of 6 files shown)


consideRatio:
@jacobtomlinson With us both being very low on time, I'll go for a judgment call and self-merge this without further review to use the limited time available to allocate as best as possible.

PR #853, opened 2025-01-21
helm-chart: remove deprecated use of traefik middleware's forceSlash option

### LINKED ISSUE #852: Stop using deprecated `forceSlack` in traefik middleware resources
I tried to stop using the `forceSlash` option in a traefik middleware, but it led to failures in k8s 1.32. See https://github.com/dask/dask-gateway/pull/851 for details on my failed attempt.

If possible, we should migrate away from using `forceSlash` since its deprecated, but its not important I figure.

gardleopard:
forceSlash is deprecated by traefik.

I ran the tests on my github user and the k8s v1.32 tests passed. Ref https://github.com/gardleopard/dask-gateway/pull/1/checks . I guess pr #851 experienced flaky tests

closes #852

### DIFF
--- a/resources/helm/dask-gateway/templates/gateway/middleware.yaml
+++ b/resources/helm/dask-gateway/templates/gateway/middleware.yaml
@@ -20,5 +20,4 @@ spec:
   stripPrefix:
     prefixes:
       - '{{ .Values.gateway.prefix | trimSuffix "/" }}'
-    forceSlash: false
 {{- end }}


consideRatio:
I'm hesitant because I'm not fully confident we don't cause issues here.

I'll try to become confident...

1. This change only impacts the helm-chart if `gateway.prefix` is not `/`
2. The helm chart has a default value of `/`
   ```yaml
   gateway:
     # Path prefix to serve dask-gateway api requests under
     # This prefix will be added to all routes the gateway manages
     # in the traefik proxy.
     prefix: /
   ```
3. Our tests typically configure this:
   ```yaml
   gateway:
   ```
4. `The forceSlash option ensures the resulting stripped path is not the empty string, by replacing it with / when necessary.`
   In other words, this is only meant to impact a situation if you'd configure `gateway.prefix` explicitly to be the empty string or null in the helm chart. Can this happen? Hmmm...

Let's go with it, I think it's probably ok. It's also not obvious if this option is defaulting to false or true even though it's documented to default to true, it seems that it may default to false in the code.

Issue #855, opened 2025-01-22
Make traefik CRD installation flaggable


gardleopard:
I'd like to make it possible to disable the CRD installation for traefik in the helmchart. This is for installation in clusters that already has traefik installed.

Is this a feature that would be accepted?

jacobtomlinson:
Sounds reasonable to me. Is there any way to catch and warn about cases where folks disable it, but it's not installed already?

gardleopard:
AFAIK it will break the helm install, as the cluster will not have the CRD definition that we try to upload.

consideRatio:
There are various strategies to provide CRDs for heln charts i think.

1. By putting them in a crds folder (current situation)
   this leads to a one-off installation as part of the chart install (not upgrade), making users need to manually update the CRDs
2. By rendering them as other manifests part of the helm chart templates

I'm not confident on what makes sense etc, but the key is to overview if putting them as templates is relevant for existing users and/or new users, and how people should migrate etc between strategies.

consideRatio:
Can one opt-out of installing a charts CRDs in the crds folder, using a helm CLI flag? If so, one could first opt out during the initial install, and then manually apply yhe dask-gateway specific CRDs as a workaround if better options aren't available for us.

gardleopard:
I will make a PR with the ability to turn on and off traefik and can test how it works etc:) It will take a few days before I can do it.

gardleopard:
I made a PR to make traefik installation optional. With the built in mechanisms of helm with regards to skipping installation of CRDs, I did not find a nice solution for opting out of traefik CRD installation while keeping the installation of the `daskclusters.gateway.dask.org`CRD.

consideRatio:
I'll go for a close of this based on #857, because I don't see a clear additional action point here besides what was done there. Thank you @gardleopard!!

PR #857, opened 2025-01-31
helm-chart: add `traefik.installTraefik` config to opt-out when traefik is installed separately

gardleopard:
This makes traefik installation optional.

Installation of CRDs happens if traefik is set to not be installed as well. Helm has built in functionality for not installing CRDs with --skip-crds, however for this project that is suboptimal, because it also blocks the install of `daskclusters.gateway.dask.org` CRD.

Relates to #855


ohait (approved):

consideRatio:
I figure the way to use dask-gateway with a standalone traefik install, one would install the dask-gateway specific CRDs manually, and then do `helm install .... --skip-crd`.

```bash
# when installing dask-gateway without traefik
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/dask/dask-gateway/<VERSION HERE>/resources/helm/dask-gateway/crds/daskclusters.yaml
helm install .... --skip-crd

# when at a later time upgrading without dask-gateway's traefik
kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/dask/dask-gateway/<VERSION HERE>/resources/helm/dask-gateway/crds/daskclusters.yaml
helm upgrade ....
```

The change in this PR seems fine and could be merged without docs as well, but could you skim through docs to see if there is a place where it's relevant to mention installing or not installing traefik @gardleopard?

.github/workflows/test.yaml
```diff
@@ -268,6 +268,89 @@ jobs:
...
+      - name: helm install
+        run: |
+          helm install \
+              test-dask-gateway \
+              resources/helm/dask-gateway \
+              --values=resources/helm/testing/chart-install-values.yaml \
+              --set traefik.installTraefik=false \
+              --set gateway.backend.namespace=default \
+              --wait \
+              --timeout 1m0s
```
consideRatio, line 334: I figure we should adopt the `--skip-crd` flag here and install the dask-gateway CRDs first in a standalone command, as that would be the procedure I figure we should recommend to avoid having dask-gateway interfere with traefik CRDs in any way.

resources/helm/dask-gateway/values.schema.yaml
```diff
@@ -478,6 +478,7 @@ properties:
       - loglevel
       - dashboard
       - service
+      - installTraefik
```
consideRatio, line 481: This is often named `enabled` in Helm charts where a sub-resource is optional, could we rename it to that? What do you think?

resources/helm/dask-gateway/values.yaml
```diff
@@ -234,6 +234,9 @@ controller:
 # that is acting as a proxy for traffic towards the gateway or user created
 # DaskCluster resources.
 traefik:
+  # If traefik is already installed in the cluster, we do not need to install traefik
```
consideRatio, line 237: This commend for example doesn't clarify the CRD complexity, I figure its worth being explicit and saying that this toggle impacts the non-CRD resources for traefik deployed by this helm chart, but that the CRDs needs to be opt-out via `--skip-crd` during installation.

.github/workflows/test.yaml
gardleopard: Done in latest commit

resources/helm/dask-gateway/values.schema.yaml
gardleopard: I am not opposed to doing it that way. The reason I chose to be more specific was because we use ingressroute and middleware in the gateway templates. So it seems a bit wierd to me that we will use and depend on traefik if traefik is set to enabled: false. What do you think?

resources/helm/dask-gateway/values.yaml
gardleopard: I changed the comment here to include crd information.

gardleopard:
How the docs render 
image

resources/helm/dask-gateway/values.schema.yaml
consideRatio: Lets go for a merge, this is a detail of little importance - feel free to open a PR if you think it should be changed, I'll go for a merge!

consideRatio:
Thank you @gardleopard!!

jabbera:
For anyone who stumbles across this the appropriate settings for a default installation of traefik in a namespace called traefik is:


DASK_GATEWAY__ADDRESS: http://proxy-public/services/dask-gateway
DASK_GATEWAY__PROXY_ADDRESS: gateway://traefik.traefik.svc.cluster.local:80

jhub helm chart:
```
hub:
  consecutiveFailureLimit: 0 # Disable 5 consecutive failurs diables hub
  services:
    dask-gateway:
      url: http://traefik.traefik.svc.cluster.local
      apiToken: <snip>
```

PR #862, opened 2025-02-03
update jupyterhub user url

minrk:
use standard `/api/user` instead

`/authorizations/token/{token}` is deprecated since jupyterhub 2.0 in 2021, superseded by `/api/user` in 0.8 in 2017.

### DIFF
--- a/dask-gateway-server/dask_gateway_server/auth.py
+++ b/dask-gateway-server/dask_gateway_server/auth.py
@@ -2,7 +2,6 @@ import base64
 import os
 import time
 import uuid
-from urllib.parse import quote
 
 import aiohttp
 from aiohttp import web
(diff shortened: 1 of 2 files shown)


minrk:
Oops, not quite right. Will fix.

consideRatio (approved):
Thank you @minrk and welcome as a code contributor to dask-gateway! :heart: :tada:

dask-gateway-server/dask_gateway_server/auth.py
```diff
@@ -376,13 +375,10 @@ async def authenticate(self, request):
...
             raise unauthorized("jupyterhub")
 
-        url = "{}/authorizations/token/{}".format(
-            self.jupyterhub_api_url,
-            quote(token, safe=""),
-        )
+        url = f"{self.jupyterhub_api_url}/user"
 
         kwargs = {
-            "headers": {"Authorization": "token %s" % self.jupyterhub_api_token},
```
minrk, line 385: Notably, `jupyterhub_api_token` is entirely not needed by JupyterHubAuth. With this implementation, dask-gateway technically doesn't need to be registered as a jupyterhub service, since it makes no use of the jupyterhub API _as itself_. That would only really be needed if the gateway began applying full oauth to protect dashboard pages, which appear to currently only be protected by the unguessable cluster names.

PR #863, opened 2025-02-03
[wip] support JupyterHub access scopes

### LINKED ISSUE #829: Consider JupyterHub authentication rework
Currently the `dask-gateway` client is reading the user's `JUPYTERHUB_API_TOKEN`, which then is passed to the `dask-gateway-server`, which then confirms the jupyterhub api token is associated with a real user via the [`hub/api/authorizations/token` api](https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#/default/get_authorizations_token__token_) by using the user's api token.

## Drawbacks

1. dask-gateway clients sends away the user's api token and its used directly - its a course authentication check that gets the full permission of the user (I think) even though dask-gateway server just needed to confirm the user was allowed
2. there isn't a mechanism to provide some users but not all users access to work with dask-gateway if what dask-gateway-server only checks if the user exists or not

## Alternatives

In [this jupyterhub documentation](https://jupyterhub.readthedocs.io/en/stable/reference/services.html#implementing-your-own-authentication-with-jupyterhub) I understand that there is a way to implement authentication that ens up requiring the user to have `access:services!service=$service-name` where service name could be `dask-gateway` for example (it should be the service name configured with jupyterhub, allowing dask-gateway to interact with jupyterhub in the first place).

## Action points

Consider if its reasonable to migrate the existing authentication check to something new that requires users to have a scope of `access:services!service=$service-name`.

minrk:
Draft because I'm not sure what the defaults/backward-compatiibility should be

Most details can be found in #829 but this would allow standard and more fine-grained control of access to the gateway service for tokens. The default is currently unchanged, for maximum compatibility.

To enable this with users being able to access this service by default from their server (minimal change from current default capabilities), a deployment needs to:

1. configure dask-gateway to know its service name (usually `dask-gateway`)
2. grant users the `access:services!service=dask-gateway` scope via the default `user` role (or a different role, if it should be less than all users, which is part of the point)
3. grant server tokens the same scope via the `Spawner.server_token_scopes` (or the `server` role, prior to JupyterHub 4.0)

The main idea of this is it enables both:

1. not all users have access to the gateway, and/or
2. not all _tokens_ have access to the gateway

since as it is now, JupyterHubAuth grants any token provided to any jupyterhub service full access to the dask gateway.

TODO:

- [ ] decide on enabling by default
- [ ] docs for change, how to enable
- [ ] update tests
- [ ] update gateway chart? Maybe also daskhub chart?

closes #829


minrk:
Also possibly relevant, since I don't know how dask-labextension works: Does the lab extension use the PageConfig.token to make requests directly to the Gateway, or does it make requests only to a server extension, which then uses $JUPYTERHUB_API_TOKEN? If it uses the jupyterlab token, the service access permission would _also_ need to be granted to `Spawner.oauth_client_allowed_scopes`

yuvipanda:
Great to see this existing :)

Given the general lack of activity in this project, I would suggest making this non-breaking to reduce potential support burden on the existing limited set of maintainers

consideRatio (commented):
Beautiful @minrk! Thank you for working this!!

I look for clarity about the one review comment I made, but I figure we can proceed with a merge besides that.

resources/helm/dask-gateway/templates/gateway/configmap.yaml
```diff
@@ -104,6 +104,9 @@ data:
                     "your config file"
                 )
         c.DaskGateway.JupyterHubAuthenticator.jupyterhub_api_url = api_url
+        service_name = get_property("gateway.auth.jupyterhub.jupyterhubServiceName")
+        if service_name is not None:
+          c.DaskGateway.JupyterHubAuthenticator.jupyterhub_service_name = service_name
```
consideRatio:
```suggestion
       c.JupyterHubAuthenticator.jupyterhub_api_url = api_url
        service_name = get_property("gateway.auth.jupyterhub.jupyterhubServiceName")
        if service_name is not None:
          c.JupyterHubAuthenticator.jupyterhub_service_name = service_name
```

PR #883, opened 2025-02-10
Fix typo in SLURM backend.

amanning9:
This typo prevents a cluster being created when a SLURM qos is set.

Like #603 but for qos instead. This list needs to be appended to instead of extended.

Apologies for not noticing this when I did #603 - we've just started using qos-s and ran into this immediately!


consideRatio (approved):

consideRatio:
Thank you @amanning9!!

PR #896, opened 2025-07-08
Update Authentication Documentation JupyterHub variable for url/token

sunnydean:


jacobtomlinson:
Did this change somewhere on the JupterHub side? Can you link to an issue or PR where this configuration option was changed?

PR #941, opened 2026-07-20
ci: test Kerberos authenticator without Hadoop

### LINKED ISSUE #924: ci: no longer testing kerberos authenticator
By disabling hadoop/yarn tests in #923, we also stopped testing kerberos authenticator, as we no longer have a ci environment where its setup and used.

I'm personally not motivated enough to allocate my time towards maintenance of this though.

GorML:
Closes #924

We use dask-gateway with Kerberos authentication at my company, so I decided
to try bringing this test coverage back.

Since #923 removed the Hadoop/YARN backend from CI, `test_kerberos_auth` has
been skipped unconditionally: it was gated on `TEST_DASK_GATEWAY_YARN` and
relied on the KDC, hostname and keytab of the old Hadoop container.

This PR restores real CI coverage of the Kerberos authenticator without
bringing back the unsupported Hadoop/YARN image:

- The test now creates its own ephemeral MIT Kerberos realm with
  [k5test](https://pypi.org/project/k5test/) (a small test-only library from
  the python-gssapi org), so no external KDC is needed. It is gated on a new
  `TEST_DASK_GATEWAY_KERBEROS` env var, and when that is set there is no
  skip path — missing dependencies fail the job instead of silently
  skipping the test.
- A new `kerberos-tests` CI job installs MIT Kerberos from apt, builds
  pykerberos via the `[kerberos]` extras, and runs the test (~2 minutes).
- Comments in `tests/requirements.txt` updated to match.

No production code changes.

Some verification beyond CI being green here:

- 30 back-to-back runs in a clean Ubuntu 24.04 container, no flakiness.
- Breaking the `HTTP/<hostname>` service principal on purpose makes the test
  fail with "Server not found in Kerberos database", so the test really
  exercises the handshake and can't pass by accident.
- The kerberos job passed 3/3 times on GitHub-hosted runners in my fork.
- Without kerberos packages and the env var the test still skips cleanly, so
  the regular test matrix is unaffected.

Unrelated notes:

- The `local backend` and `pbs`/`slurm backend` jobs currently fail on
  `main` for reasons unrelated to this PR (missing `dask-scheduler` CLI with
  latest distributed; pillow sdist build without gcc in the CI containers).
  Happy to open separate issues for those.
- `continuous_integration/docker/hadoop/` is orphaned since #923 and could
  be removed in a follow-up.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -334,3 +334,59 @@ jobs:
       - name: Print logs for debugging
         if: always()
         run: ./continuous_integration/docker/${{ matrix.backend }}/print_logs.sh
+
+  # This job tests the Kerberos authenticator against an ephemeral MIT
+  # Kerberos realm created with k5test directly on the runner, without needing
+  # a containerized backend.
+  #
+  # ref: https://github.com/dask/dask-gateway/issues/924
+  #
+  kerberos-tests:
+    name: "kerberos authentication"
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+      - uses: actions/setup-go@v6
+        with:
+          go-version: "1.25"
+          cache-dependency-path: "**/*.sum"
+
+      # krb5-kdc and krb5-admin-server provide the KDC that k5test manages,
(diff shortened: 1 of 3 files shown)


consideRatio:
Thank you for adressing this, we have another ci failure on main branch - so the failures may very well be unrelated.

This PR seems very reasonable though

